PR #477, opened 2021-08-04
Add doc comments for authentication related records  

aneeshafedo:
## Purpose
1. Update client init function return statement doc comment to `return - An error if connector initialization failed`
2. Throw an error when API Key, OAuth 2.0 and Http authentication mechanisms available at the same time
3. Add documentation comments for config records

> Resolves https://github.com/ballerina-platform/ballerina-openapi/issues/476

## Security checks


openapi-cli/src/main/java/io/ballerina/openapi/generators/client/BallerinaAuthConfigGenerator.java
```diff
@@ -131,15 +141,17 @@
     private boolean isHttpOROAuth = false;
     private final Set<String> authTypes = new LinkedHashSet<>();
     private String apiKeyDescription;
+    private final GeneratorUtils generatorUtils;
```
lnash94, line 144: Shall we remove adding this util class here and keep the methods static since the function `createParamAPIDoc` behave like the util?

aneeshafedo: yes, but the only usage of  that function is only in this class. What are the drawbacks of creating an `GeneratorUtil` class instance here ?

lnash94: We can keep the consistency of the Util class as it is instead of creating the instance of the class. :)

aneeshafedo: Yeah sure we can do that. The way I have used here is the same way we have done it in the `BallerinaClientGenerator`. Are u suggesting to apply the change to any class that have used `createParamAPIDoc` function ?

lnash94: Yeah, it is better to remove the GeneratorUtil instance usage of `BallerinaClientGenerator` and ` FunctionSignatureGenerator`, Make it as a Util, if it is hard ftm, let's do it in the next fixes.

aneeshafedo: Will address this suggestion in the next PR

indikasampath2000 (approved):

PR #539, opened 2021-08-24
[Client] Bug fix in record field generation when the type of inline allOf schema in not specified 

aneeshafedo:
In the current connector generation, record fields are not generated in below scenario

```
Pet:
      allOf:
        - properties:
            id:
              type: integer
              format: int64
            name:
              type: string
            tag:
              type: string
            type:
              type: string
        - properties:
            entries:
              type: array
              items:
                type: string
```
The issue is due to not specifying the type in the schemas under allOf schema.  At the times that the type of the schema is not specified but the properties are given, the schema is type: object by default

Resolves https://github.com/ballerina-platform/ballerina-openapi/issues/370

## Goals

In the above scenario. ballerina code should be generated as below. 
```ballerina
public type Pet record {
    int id?;
    string name?;
    string tag?;
    string 'type?;
    string[] entries?;
};
```

## Security checks

### DIFF
--- a/openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/BallerinaSchemaGenerator.java
+++ b/openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/BallerinaSchemaGenerator.java
@@ -356,7 +356,7 @@ public class BallerinaSchemaGenerator {
             if (allOfSchema.getType() == null && allOfSchema.get$ref() != null) {
                 //Generate typeReferenceNode
                 getAllOfRecordFieldForReference(schemaDoc, recordFieldList, allOfSchema);
-            } else if (allOfSchema instanceof ObjectSchema && (allOfSchema.getProperties() != null)) {
+            } else if (allOfSchema.getProperties() != null) {
                 if (allOfSchema.getDescription() != null) {
                     schemaDoc.addAll(DocCommentsGenerator.createAPIDescriptionDoc(
                             allOfSchema.getDescription(), false));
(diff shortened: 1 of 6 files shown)


openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/BallerinaSchemaGenerator.java
```diff
@@ -356,7 +356,7 @@ private RecordTypeDescriptorNode getAllOfRecordTypeDescriptorNode(OpenAPI openAp
             if (allOfSchema.getType() == null && allOfSchema.get$ref() != null) {
                 //Generate typeReferenceNode
                 getAllOfRecordFieldForReference(schemaDoc, recordFieldList, allOfSchema);
-            } else if (allOfSchema instanceof ObjectSchema && (allOfSchema.getProperties() != null)) {
+            } else if (allOfSchema instanceof ObjectSchema || allOfSchema.getProperties() != null) {
```
lnash94, line 359: Won't this `||`  condition be lead to an NPE?

aneeshafedo: yes there is a possibility of getting a NPE if user has defined any Object schema without any properties. 
```
allOf:
        - description: ""
          type: object
        - properties:
            entries:
              type: array
              items:
                type: string
```

indikasampath2000 (approved):

PR #564, opened 2021-09-06
Fix initial implementation for Ballerina to OAS tool to support caching configuration in response using `@http:CacheConfig` annotation

lnash94:
## Automation tests
 - Unit tests  - add
   > Code coverage information
 - Integration tests

## Security checks


openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
```diff
@@ -0,0 +1,164 @@
...
+    private boolean proxyRevalidate = false;
+    private int maxAge = -1;
+    private int sMaxAge = -1;
+    private List<String> noCacheFields = new ArrayList<>();
+    private List<String>  privateFields = new ArrayList<>();
+    private boolean setETag = true;
+    private boolean setLastModified = true;
+
+    public CacheConfigAnnotation() {
+
```
shafreenAnfar, line 42: Let's remove all the newlines in this class.

lnash94: Fix with this commit https://github.com/ballerina-platform/ballerina-openapi/pull/564/commits/05335d11c3cda4bfd7159122d2c3b02161a300ad

shafreenAnfar (approved):

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.ArrayList;
+import java.util.List;
+/**
+ * This class uses to store all the cache configuration details.
+ *
+ * @since 2.0.0
+ */
+public class CacheConfigAnnotation {
+    private boolean mustRevalidate = false;
```
TharmiganK, line 28: In the default `@http:CacheConfig`, we have enabled the `must-revalidate` directive as well. So can we change this to `true`

openapi-cli/src/test/resources/ballerina-to-openapi/expected_gen/response/cache_config_03.yaml
```diff
@@ -0,0 +1,23 @@
...
+    get:
+      operationId: operation_get_/cachingBackEnd
+      responses:
+        "200":
+          description: Ok
+          headers:
+            Cache-Control:
+              schema:
+                type: string
+                default: "no-cache=\"field03,fields04\",private=\"field1,filed2\"\,max-age=5"
```
TharmiganK, line 12: There is an additional escape character before `,max-age`

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
```diff
@@ -0,0 +1,139 @@
...
+    private boolean noCache = false;
+    private boolean noStore = false;
+    private boolean noTransform = false;
+    private boolean isPrivate = false;
+    private boolean proxyRevalidate = false;
+    private int maxAge = -1;
```
TharmiganK, line 34: Also we set `max-age=3600` as default

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
lnash94: Sure let's change with it :). There is a concern about this if some user doesn't set must-validate in their cache-config ex:
```ballerina 
resource function get cachingBackEnd(http:Request req) returns @http:CacheConfig{maxAge : 5} string {
       return "Hello, World!!";}
```
what would be a default string value we get ? is it similar to this `public, max-age=5` or `public, max-age=5, must-revalidate` ?

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
lnash94: The same concern there as well, if the user doesn't provide any max-age value we need to add a default value to the generated string?
If in case the user set the negative value there could you please let me know whether we ignore it by adding the max-age field to the string or are we add `max-age` with the default value?

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
TharmiganK: It will be `public, max-age=5, must-revalidate`

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
TharmiganK: **Case 1 :** If the `max-age` is not set in the annotation, then we add the **default** value `max-age=3600`.

**Case 2 :** If the `max-age` is set to a negative value, then we **do not** add `max-age` as a directive (ignore it in the string). This is how the user can disable the `max-age` directive.

**Case 3 :** If the `max-age` is set to a positive value, then we add that value as `max-age`.

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
lnash94: Thank you , will change with this :)

lnash94: Fix with https://github.com/ballerina-platform/ballerina-openapi/pull/564/commits/880d1122b2fc897bb656d1f44c31bdf3a7906695

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/CacheConfigAnnotation.java
lnash94: Fix with https://github.com/ballerina-platform/ballerina-openapi/pull/564/commits/880d1122b2fc897bb656d1f44c31bdf3a7906695

PR #566, opened 2021-09-07
Improve OpenAPI compiler plugin

ayeshLK:
Following improvements should be done:
1. Try not to create intermediate `resources` directory and just keep the generated OpenAPI spec in-memory and use it to update generated `jar` files. (if possible)
2. Intermediate `resources` directory should not be created inside `target/bin`
3. Generated `resources` should be packed into generated `thin-jar` and then it will be packed into `executable-jar`.
4. Add compiler warnings for plugin failures.
5. Include a test case to check how this works for dockerized HTTP service.

Fixes [#1854](https://github.com/ballerina-platform/ballerina-standard-library/issues/1854)


lnash94:
LGTM

lnash94 (approved):

azinneera (approved):

PR #647, opened 2021-10-08
[Validator] Remove usage of deprecated semanticAPI

lnash94:

### DIFF
--- a/openapi-validator/src/main/java/io/ballerina/openapi/validator/TypeSymbolToJsonValidatorUtil.java
+++ b/openapi-validator/src/main/java/io/ballerina/openapi/validator/TypeSymbolToJsonValidatorUtil.java
@@ -104,7 +104,7 @@ public class TypeSymbolToJsonValidatorUtil {
                 TypeSymbol recordType = null;
                 isExitType = true;
                 Optional<TypeSymbol> symbol = semanticModel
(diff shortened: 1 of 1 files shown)


openapi-validator/src/main/java/io/ballerina/openapi/validator/TypeSymbolToJsonValidatorUtil.java
```diff
@@ -194,14 +194,15 @@
                         TypeMismatch validationError = new TypeMismatch(fieldSymbol.getValue().getName().orElseThrow(),
                                 convertTypeToEnum(entry.getValue().getType()),
                                 convertTypeToEnum(fieldSymbol.getValue().typeDescriptor().typeKind().getName()),
-                                componentName, fieldSymbol.getValue().location());
+                                componentName, fieldSymbol.getValue().getLocation().get());
```
NipunaRanasinghe, line 197: Don't we need to check whether the location is empty? 

@pubudu91 In which contexts can we expect that the location being empty?

pubudu91: Location would be empty for type symbols in general (except `TypeReferenceTypeSymbol`)

PR #720, opened 2021-11-12
Add command-line option to generate test cases

SanduDS:
## Purpose
Added "--with-tests" option to add test files to the generated code.
Fixes https://github.com/ballerina-platform/ballerina-openapi/issues/673

## Approach
Added new option "--with-tests"

## Automation tests
 - Unit tests 
   > Done
 - Integration tests
   > Done

## Security checks

### DIFF
--- a/openapi-cli/src/main/java/io/ballerina/openapi/cmd/OpenApiCmd.java
+++ b/openapi-cli/src/main/java/io/ballerina/openapi/cmd/OpenApiCmd.java
@@ -95,6 +95,9 @@ public class OpenApiCmd implements BLauncherCmd {
     @CommandLine.Option(names = {"--json"}, description = "Generate json file")
     private boolean generatedFileType;
 
+    @CommandLine.Option(names = {"--with-tests"}, hidden = true, description = "Generate test files")
+    private boolean includeTestFiles;
+
     @CommandLine.Parameters
     private List<String> argList;
 
@@ -219,6 +222,7 @@ public class OpenApiCmd implements BLauncherCmd {
     private void openApiToBallerina(String fileName, Filter filter) throws IOException {
         CodeGenerator generator = new CodeGenerator();
         generator.setLicenseHeader(this.setLicenseHeader());
+        generator.setIncludeTestFiles(this.includeTestFiles);
         final File openApiFile = new File(fileName);
         String serviceName;
         if (generatedServiceName != null) {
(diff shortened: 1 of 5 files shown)


openapi-cli/src/main/java/io/ballerina/openapi/cmd/OpenApiCmd.java
```diff
@@ -306,6 +310,7 @@ private String setLicenseHeader() {
      */
     private void generatesClientFile(CodeGenerator generator, String clientName, Path resourcePath, Filter filter) {
         try {
+            generator.setIncludeTestFiles(includeTestFiles);
```
lnash94, line 313: Is this essential to set a test flag `setIncludeTestFiles(includeTestFiles)` since it is already set in line 225 at the main command call?

openapi-cli/src/test/java/io/ballerina/openapi/CodeGeneratorTest.java
```diff
@@ -90,6 +91,7 @@ public void generateClient() {
         CodeGenerator generator = new CodeGenerator();
         try {
             String expectedClientContent = getStringFromGivenBalFile(expectedServiceFile, "generate_client.bal");
+            generator.setIncludeTestFiles(true);
```
lnash94, line 94: Shall we add separate tests for `--with-test` flag for service generation, client generation, both service and client generation scenarios instead of changing already existing tests 🙂 it will be more clear for filter out the test with flag in command WDYT?

aneeshafedo: +1. Adding a separate test case for --with-test scenario will be more clear.

aneeshafedo (commented):
Other than the test case in `CodeGeneratorTests` there are few `test.bal` generation related tests [here](https://github.com/ballerina-platform/ballerina-openapi/blob/master/openapi-cli/src/test/java/io/ballerina/openapi/generators/testcases/BallerinaTestGeneratorTests.java).  

Please do necessary changes to meaningfully execute `BallerinaTestGeneratorTests`.

openapi-cli/src/test/java/io/ballerina/openapi/CodeGeneratorTest.java
SanduDS: There is a separate test case for --with-test but here what we have done is that enabling the addition of tests/test.bal and generation other test-related files as the function was designed earlier, without changing the expected output of the test cases.

aneeshafedo (approved):

LakshanSS (approved):

lnash94 (approved):

PR #722, opened 2021-11-12
Update LS extension response type

gigara:
From this PR, the YAML content's data type in the response of LS extension is refactored into a JSON array.

### DIFF
--- a/openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterResponse.java
+++ b/openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterResponse.java
@@ -18,9 +18,8 @@ package io.ballerina.openapi.extension;
  *  under the License.
  */
 
-import io.ballerina.openapi.converter.service.OASResult;
+import com.google.gson.JsonArray;
 
-import java.util.List;
 import java.util.Optional;
 
 /**
@@ -32,7 +31,8 @@ public class OpenAPIConverterResponse {
     @Deprecated
     private String yamlContent;
 
-    private List<OASResult> content;
+    // Json Array of OASResults
+    private JsonArray content;
     private String error;
 
     public OpenAPIConverterResponse() {
@@ -48,11 +48,11 @@ public class OpenAPIConverterResponse {
         this.yamlContent = yamlContent;
     }
 
-    public List<OASResult> getContent() {
+    public JsonArray getContent() {
         return content;
     }
 
-    public void setContent(List<OASResult> content) {
+    public void setContent(JsonArray content) {
         this.content = content;
     }
 
--- a/openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
+++ b/openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
@@ -19,10 +19,21 @@
 
 package io.ballerina.openapi.extension;
 
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import io.ballerina.compiler.api.SemanticModel;
 import io.ballerina.compiler.syntax.tree.SyntaxTree;
+import io.ballerina.openapi.converter.diagnostic.OpenAPIConverterDiagnostic;
 import io.ballerina.openapi.converter.service.OASResult;
 import io.ballerina.openapi.converter.utils.ServiceToOpenAPIConverterUtils;
+import io.ballerina.projects.Document;
+import io.ballerina.projects.DocumentId;
+import io.ballerina.projects.Module;
+import io.ballerina.tools.diagnostics.Location;
 import org.ballerinalang.annotation.JavaSPIService;
 import org.ballerinalang.langserver.commons.service.spi.ExtendedLanguageServerService;
 import org.ballerinalang.langserver.commons.workspace.WorkspaceManager;
@@ -52,13 +63,6 @@ public class OpenAPIConverterService implements ExtendedLanguageServerService {
     @Override
     public void init(LanguageServer langServer, WorkspaceManager workspaceManager) {
         this.workspaceManager = workspaceManager;
-    }
-
-    public OpenAPIConverterService(WorkspaceManager workspaceManager) {
-        this.workspaceManager = workspaceManager;
-    }
-
-    public OpenAPIConverterService() {
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +104,93 @@ public OpenAPIConverterService() {
...
+            if (module.isEmpty()) {
+                response.setError("Error while getting the module.");
+                return response;
             }
+            Module defaultModule = module.get();
+            JsonArray specs = new JsonArray();
+            for (DocumentId currentDocumentID : defaultModule.documentIds()) {
+                Document document = defaultModule.document(currentDocumentID);
+
+                SyntaxTree syntaxTree = document.syntaxTree();
```
lnash94, line 131: Why can't we directly use ` workspaceManager.syntaxTree()` to get syntax tree instead of going through the module :) ?

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +104,93 @@ public OpenAPIConverterService() {
...
+
+                generateServiceJson(response, document.name(), oasResults, specs);
+            }
+            response.setContent(specs);
             return response;
         });
     }
 
+    /**
+     * Generate openAPI json for a service.
```
lnash94, line 143: Shall we add a new line below and apply to other relevant places?

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterResponse.java
```diff
@@ -32,7 +31,8 @@
     @Deprecated
     private String yamlContent;
 
-    private List<OASResult> content;
+    // Json Array of OASResults
+    private JsonArray content;
```
prabushi, line 35: Would it be okay to change the response type without deprecating the existing API or creating a new API? This will be a breaking change for the existing users.

gigara: @NipunaMarcus confirmed that there is no current usage in this API and we can change that.

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +104,95 @@ public OpenAPIConverterService() {
...
-                response.setError(errorString.toString());
-            } else {
-                response.setError(null);
-                List<OASResult> oasResult = ServiceToOpenAPIConverterUtils.generateOAS3Definition(syntaxTree.get(),
-                        semanticModel.get(), null, false, null);
-                //Response handle with returning list of {@code OASResult} model.
-                response.setContent(oasResult);
+            Path filePath = Path.of(request.getDocumentFilePath());
+            Optional<SemanticModel> semanticModel = workspaceManager.semanticModel(filePath);
+            Optional<Module> module = workspaceManager.module(filePath);
```
prabushi, line 116: Does this always return the default module? According to the API description, it returns the module in the given path. If so, when the path is given for a ballerina file in another module (other than default), it'll return that specific module. Can you check on this?

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +104,95 @@ public OpenAPIConverterService() {
...
+     * @param response OpenAPIConverterResponse to set error message
+     * @param documentName Ballerina document name
+     * @param oasResults OAS Results list
+     * @param specs Specs array to add the service
+     */
+    private void generateServiceJson(OpenAPIConverterResponse response, String documentName,
+                                     List<OASResult> oasResults, JsonArray specs) {
+        for (OASResult oasResult : oasResults) {
+            if (oasResult.getOpenAPI().isPresent()) {
+                if (oasResult.getJson().isPresent()) {
```
prabushi, line 154: Shall we check isEmpty() and continue early?

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +104,95 @@ public OpenAPIConverterService() {
...
+                    JsonObject spec = new JsonObject();
+                    JsonElement json = JsonParser.parseString(oasResult.getJson().get()).getAsJsonObject();
+                    JsonArray diagnosticsJson = getDiagnosticsJson(oasResult);
+
+                    spec.addProperty("serviceName",
+                            documentName + " - " + oasResult.getOpenAPI().get().getInfo().getTitle());
+                    spec.add("spec", json);
+                    spec.add("diagnostics", diagnosticsJson);
```
prabushi, line 162: Shall we define constants?

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +104,95 @@ public OpenAPIConverterService() {
...
+     * @param oasResults OAS Results list
+     * @param specs Specs array to add the service
+     */
+    private void generateServiceJson(OpenAPIConverterResponse response, String documentName,
+                                     List<OASResult> oasResults, JsonArray specs) {
+        for (OASResult oasResult : oasResults) {
+            if (oasResult.getOpenAPI().isPresent()) {
+                if (oasResult.getJson().isPresent()) {
+                    JsonObject spec = new JsonObject();
+                    JsonElement json = JsonParser.parseString(oasResult.getJson().get()).getAsJsonObject();
```
prabushi, line 156: Let's rename the `getJson()` method to a more meaningful name.

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
gigara: Fixed

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
gigara: Fixed

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
gigara: Fixed

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
gigara: Cannot rename because of the backward incompatibility.

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterResponse.java
NipunaMarcus: currently this is been used only with beta3 ... but this will be used mostly only for this going forward hence I think we don't have to maintain this... but yeah we can deprecate the old and give add a new API but I don't see a point doing that though ...

nadeeshaan: What's the reason for converting the return type to a `JsonArray`? This is too generic in the API point of view, in the sense we expose an API which we can return a JsonArray of any type of elements. Also, isn't the IIRC, the JSON-RPC layer converts the list to a JsonArray under the hood.

nadeeshaan (changes requested):

gigara:
@nadeeshaan The JSON-RPC layer failed to convert this list and threw and exception before. Because of that we changed it to a json.

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterResponse.java
NipunaMarcus: there was a serialization issue with the type used in List nadeeshaan... it cannot serialize this List hence we decided to to go with a json Array

gigara: The JSON-RPC layer failed to convert this list and threw an exception before. Because of that we changed it to a json.

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +112,96 @@ public OpenAPIConverterService() {
...
+
+            if (semanticModel.isEmpty()) {
+                response.setError("Error while generating semantic model.");
+                return response;
+            }
+            if (module.isEmpty()) {
+                response.setError("Error while getting the module.");
+                return response;
+            }
+            Module defaultModule = module.get().currentPackage().getDefaultModule();
```
nadeeshaan, line 134: Are we considering the default module only?

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
```diff
@@ -100,30 +112,96 @@ public OpenAPIConverterService() {
...
+        for (OpenAPIConverterDiagnostic diagnostic : diagnostics) {
+            JsonObject diagnosticJson = new JsonObject();
+            diagnosticJson.addProperty(MESSAGE, diagnostic.getMessage());
+            diagnosticJson.addProperty(SEVERITY, diagnostic.getDiagnosticSeverity().name());
+            Optional<Location> diagnosticLocation = diagnostic.getLocation();
+            if (diagnosticLocation.isPresent()) {
+                GsonBuilder builder = new GsonBuilder();
+                builder.serializeNulls();
+                Gson gson = builder.create();
+                diagnosticJson.add(LOCATION, gson.toJsonTree(diagnosticLocation.get().lineRange()));
```
nadeeshaan, line 198: What are the nullable fields which can be in the `LineRange`?

nadeeshaan (changes requested):

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
gigara: Yes. Because services can only be in the default module AFAIK.

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterResponse.java
nadeeshaan: I think we need to figure out what's the actual serialisation issue first. And then we can expose a proper data model, may be a new data model as the response.

gigara: As we discussed offline, this can be considered as a special use case that we need to pass the data as an Array since the third party library doesn't support the serialization.

openapi-ls-extension/src/main/java/io/ballerina/openapi/extension/OpenAPIConverterService.java
lnash94: There are situations with handling some exceptions during the code generator process in the tool since it is hard to catch the location for exceptions we pass the `null` as value for the location to the diagnostic `OpenAPIConverterDiagnostic` model.

nadeeshaan (approved):
LGTM based on the comments

PR #926, opened 2022-03-14
[BallerinaToOpenAPI]Fix mapping to `application/x-www-form-urlencoded` when it has response and request body in resources

lnash94:
Fix  ballerina-platform/ballerina-standard-library#5391 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIRequestBodyMapper.java
```diff
@@ -242,31 +241,14 @@ private void handleArrayTypePayload(Map<String, Schema> schema, ArrayTypeDescrip
     private void handleMultipleMIMETypes(RequestBody bodyParameter,
                                          SeparatedNodeList<MappingFieldNode> fields,
                                          RequiredParameterNode payloadNode, Map<String, Schema> schema) {
+        List<String> mimeTypes = new ArrayList<>();
+        for (AnnotationNode annotation : payloadNode.annotations()) {
+            mimeTypes = extractAnnotationFieldDetails(HTTP_PAYLOAD, MEDIA_TYPE, annotation, semanticModel);
```
NipunaRanasinghe, line 246: Why are we reassigning values for `memeTypes` inside the for-loop in here? when we have multiple annotation, shouldn't we append the annotation data instead?

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIResponseMapper.java
```diff
@@ -478,6 +480,19 @@ private void setCacheConfigField(CacheConfigAnnotation cacheConfig, String field
             case OPTIONAL_TYPE_DESC:
                 return getAPIResponses(operationAdaptor, apiResponses,
                         ((OptionalTypeDescriptorNode) typeNode).typeDescriptor(), customMediaPrefix, headers);
+            case MAP_TYPE_DESC:
+                apiResponse = setCacheHeader(headers, apiResponse, HTTP_200);
```
NipunaRanasinghe, line 484: Shouldn't we ideally implement these type of functionalities as setter instead of util functions? ( i.e. `apiResponse.setCacheHeader()` will be a better implementation from the OOP point of view)

openapi-cli/src/test/resources/ballerina-to-openapi/expected_gen/response/rs_with_url_encode.yaml
```diff
@@ -22,4 +22,41 @@ paths:
...
+          description: Ok
+          content:
+            application/x-www-form-urlencoded:
+              schema:
+                type: object
+                additionalProperties:
+                  type: string
+  /baa:
+    get:
+      operationId: operation_get_/baa
```
NipunaRanasinghe, line 39:
```suggestion
      operationId: operation_get_/bar
```

openapi-cli/src/test/resources/ballerina-to-openapi/expected_gen/response/rs_with_url_encode.yaml
```diff
@@ -22,4 +22,41 @@ paths:
...
+          description: Ok
+          content:
+            application/json:
+              schema:
+                type: object
+                additionalProperties:
+                  type: string
+  /baaint:
+    get:
+      operationId: operation_get_/baaint
```
NipunaRanasinghe, line 51:
```suggestion
      operationId: operation_get_/barint
```

openapi-cli/src/test/resources/ballerina-to-openapi/response/rs_with_url_encode.bal
```diff
@@ -18,4 +18,25 @@ service /payloadV on helloEp {
...
         };
     }
+    resource function get foo() returns @http:Payload {mediaType: mime:APPLICATION_FORM_URLENCODED} map<string> {
+        map<string> ms = {
+            "x": "abc",
+            "y": "cdf"
+        };
+        return ms;
+    }
+    resource function get baa() returns @http:Payload map<string> {
```
NipunaRanasinghe, line 28:
```suggestion
    resource function get bar() returns @http:Payload map<string> {
```

openapi-cli/src/test/resources/ballerina-to-openapi/response/rs_with_url_encode.bal
```diff
@@ -18,4 +18,25 @@ service /payloadV on helloEp {
...
+        map<string> ms = {
+            "x": "abc",
+            "y": "cdf"
+        };
+        return ms;
+    }
+    resource function get baaint() returns @http:Payload map<int> {
```
NipunaRanasinghe, line 35:
```suggestion
    resource function get barint() returns @http:Payload map<int> {
```

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIResponseMapper.java
lnash94: `apiResponse` object is created using the `ApiResponse` which is implemented using the swagger model. since it is a dependency model implemented one, we won't able to define a specific `setCacheHeader()` function. 🙁

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIRequestBodyMapper.java
lnash94: Annotation in metadata is an array list. It can be service annotation, a user defines annotation so on. inside the for loop, we extract the annotation details and get the list of mimeTypes.  for avoiding repeating annotation check I moved this implement to its parent function 🙂

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIResponseMapper.java
NipunaRanasinghe: In that case maybe we can implement proxy design pattern on the swagger data model to write our own setter functions :) but since that refactoring effort will take some time, lets keep them this way for now

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIRequestBodyMapper.java
```diff
@@ -123,8 +121,16 @@ public void handlePayloadAnnotation(RequiredParameterNode payloadNode, Map<Strin
             if (mapMime != null) {
                 fields = mapMime.fields();
             }
+
+            // Handle multiple mime type when attached with request payload annotation.
+            // ex: <pre> @http:Payload { mediaType:["application/json", "application/xml"] } Pet payload </pre>
             if (fields != null && (!fields.isEmpty() || fields.size() != 0)) {
```
NipunaRanasinghe, line 127: Why do we need to keep both `!fields.isEmpty()` and `fields.size() != 0` checks in here?

lnash94: Yeah, we can remove one condition there. we earlier added this due to an issue with API. as an example we have `[ , ]` array this is not empty but there is no item. since these issues have already been fixed from API. I will remove unnecessary checks here. 🙂

NipunaRanasinghe (approved):
LGTM

shafreenAnfar:
Went through the test-cases and looks good to me.  In addition to `map` we will soon support `records` as well.

lnash94:
Thank you, I will create a task to track support for openapi tool. :)

PR #928, opened 2022-03-20
[Schema]Fix removing openapi to ballerina array `maxItem` count generation in schema

lnash94:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/TypeGeneratorUtils.java
+++ b/openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/TypeGeneratorUtils.java
@@ -85,27 +85,28 @@ public class TypeGeneratorUtils {
      * Get SchemaType object relevant to the schema given.
      *
      * @param schemaValue Schema object
+     * @param typeNameToken parameter name
      * @return Relevant SchemaType object
      */
-    public static TypeGenerator getTypeGenerator(Schema<?> schemaValue) {
+    public static TypeGenerator getTypeGenerator(Schema<?> schemaValue, IdentifierToken typeNameToken) {
         if (schemaValue.get$ref() != null) {
-            return new ReferencedTypeGenerator(schemaValue);
+            return new ReferencedTypeGenerator(schemaValue, typeNameToken);
         } else if (schemaValue instanceof ComposedSchema) {
             ComposedSchema composedSchema = (ComposedSchema) schemaValue;
             if (composedSchema.getAllOf() != null) {
-                return new AllOfRecordTypeGenerator(schemaValue);
+                return new AllOfRecordTypeGenerator(schemaValue, typeNameToken);
             } else {
-                return new UnionTypeGenerator(schemaValue);
+                return new UnionTypeGenerator(schemaValue, typeNameToken);
             }
         } else if ((schemaValue.getType() != null && schemaValue.getType().equals(OBJECT)) ||
                 schemaValue instanceof ObjectSchema || schemaValue.getProperties() != null) {
-            return new RecordTypeGenerator(schemaValue);
+            return new RecordTypeGenerator(schemaValue, typeNameToken);
         } else if (schemaValue instanceof ArraySchema) {
-            return new ArrayTypeGenerator(schemaValue);
+            return new ArrayTypeGenerator(schemaValue, typeNameToken);
         } else if (schemaValue.getType() != null && primitiveTypeList.contains(schemaValue.getType())) {
-            return new PrimitiveTypeGenerator(schemaValue);
+            return new PrimitiveTypeGenerator(schemaValue, typeNameToken);
         } else { // when schemaValue.type == null
-            return new AnyDataTypeGenerator(schemaValue);
+            return new AnyDataTypeGenerator(schemaValue, typeNameToken);
         }
     }
     /**
@@ -148,7 +149,8 @@ public class TypeGeneratorUtils {
             List<Node> schemaDoc = getFieldApiDocs(field.getValue());
             NodeList<Node> schemaDocNodes = createNodeList(schemaDoc);
             IdentifierToken fieldName = AbstractNodeFactory.createIdentifierToken(fieldNameStr);
-            TypeDescriptorNode fieldTypeName = getTypeGenerator(field.getValue()).generateTypeDescriptorNode();
+            TypeDescriptorNode fieldTypeName = getTypeGenerator(field.getValue(), fieldName)
+                    .generateTypeDescriptorNode();
             MarkdownDocumentationNode documentationNode = createMarkdownDocumentationNode(schemaDocNodes);
             MetadataNode metadataNode = createMetadataNode(documentationNode, createEmptyNodeList());
             if (required != null) {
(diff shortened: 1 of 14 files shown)


openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/TypeGeneratorUtils.java
```diff
@@ -85,27 +85,28 @@
      * Get SchemaType object relevant to the schema given.
      *
      * @param schemaValue Schema object
+     * @param typeNameToken parameter name
      * @return Relevant SchemaType object
      */
-    public static TypeGenerator getTypeGenerator(Schema<?> schemaValue) {
+    public static TypeGenerator getTypeGenerator(Schema<?> schemaValue, IdentifierToken typeNameToken) {
```
aneeshafedo, line 91: Not getting the reason to pass `typeNameToken` along. How does it affect to array length configuration ?

lnash94: yeah, it is right, the schema is enough for the drive to get the type of field.

But this is for generating the warning message more detailed full (as per discuss offline,  to provide warning message including details we are not going to support maxItem), by looking at schema we won't able to get enough details to populate the message with property or field it occurs in spec since it only includes the type of the given field or variable. 🙁 

Therefore I pass the type name here.  I have noticed that in a few other places, we are throwing common messages due to this matter, hope this would help to refactor those messages as well with fields. 
Since this is a warning and YAML can contain more fields with maxItem, hope we won't able to go with a common message like previous approaches.

Please make me correct, if I miss the way we can catch that detail through only looking at the schema? 🙂

openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/ballerinatypegenerators/ArrayTypeGenerator.java
```diff
@@ -83,14 +84,9 @@ public TypeDescriptorNode generateTypeDescriptorNode() throws BallerinaOpenApiEx
...
-        Token arrayLengthToken = null;
         if (arraySchema.getMaxItems() != null) {
-            if (arraySchema.getMaxItems() <= MAX_ARRAY_LENGTH) {
-                arrayLengthToken = createIdentifierToken(String.valueOf(arraySchema.getMaxItems()));
-            } else {
-                throw new BallerinaOpenApiException("Maximum item count defined in the definition exceeds the " +
-                        "maximum ballerina array length.");
-            }
+            outStream.println("WARNING: Maximum item count defined in the array ' " + typeName.text() +
+                    "' property in the definition will not add to the maximum length for ballerina array.");
```
NipunaRanasinghe, line 89:
```suggestion
            outStream.println("WARNING: Maximum array item count defined for the property ' " + typeName.text() + "' in the definition will be ignored");
```

openapi-cli/src/test/java/io/ballerina/openapi/generators/schema/ArrayDataTypeTests.java
```diff
@@ -153,28 +154,8 @@ public void arrayHasOneOfItemsWithNullable() throws IOException, BallerinaOpenAp
                 "schema/ballerina/array_with_oneOf_complex.bal", syntaxTree);
     }
 
-    @Test(description = "Array schema has max items count that ballerina doesn't support",
```
NipunaRanasinghe, line 156: Shouldn't we add a negative test case and check for the warning?

openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/TypeGeneratorUtils.java
NipunaRanasinghe: @lnash94 Then why don't we pass just the `typeName` instead of populating the ST token of the type name identifier? Passing the type name when creating a type generator instance makes more sense to me. @aneeshafedo @lnash94 WDYT?

aneeshafedo: Or we can pass `Map.Entry<String, Schema> schema ` to the `TypeGenerator` without passing only the `schema.value`. Then `TypeGenerator` has access to both `schema.value` & `schema.key`. Just an idea.

openapi-cli/src/test/java/io/ballerina/openapi/generators/schema/ArrayDataTypeTests.java
lnash94: yeah, I will add test for negative tests

openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/TypeGeneratorUtils.java
lnash94: > @lnash94 Then why don't we pass just the `typeName` instead of populating the ST token of the type name identifier? Passing the type name when creating a type generator instance makes more sense to me. @aneeshafedo @lnash94 WDYT?

Here I passed the ST token by thinking in case we need this in future since we generate Syntax Tree. 🙂 

> Or we can pass `Map.Entry<String, Schema> schema ` to the `TypeGenerator` without passing only the `schema.value`. Then `TypeGenerator` has access to both `schema.value` & `schema.key`. Just an idea.
This is also nice since swagger provides a map with key-value pairs and we can pick one from them to pass one element of the map as key-value pair  🙂 

I'm afraid that if we hope to use this schema generator for other JSON and record/ type generation tasks, what would be the more sense and effective way to pass value to type Generator API, by considering these passing two parameters separately make sense to me 🤔 ? @NipunaRanasinghe @aneeshafedo  WDYT?

NipunaRanasinghe: +1 for both, but slightly prefer the 2 args constructor considering the readability and cleanliness

aneeshafedo: +1

openapi-cli/src/test/java/io/ballerina/openapi/generators/schema/ArrayDataTypeTests.java
```diff
@@ -154,6 +154,26 @@ public void arrayHasOneOfItemsWithNullable() throws IOException, BallerinaOpenAp
...
+    @Test(description = "Array schema has max items count that ballerina doesn't support, in record field",
+            expectedExceptions = BallerinaOpenApiException.class,
+            expectedExceptionsMessageRegExp = "Maximum item count defined in the definition exceeds the.*")
+    public void arrayHasMaxItemsExceedLimit02() throws IOException, BallerinaOpenApiException {
+        Path definitionPath = RES_DIR.resolve("swagger/array_exceed_max_item_02.yaml");
+        OpenAPI openAPI = codeGenerator.normalizeOpenAPI(definitionPath, true);
+        BallerinaTypesGenerator ballerinaSchemaGenerator = new BallerinaTypesGenerator(openAPI);
+        syntaxTree = ballerinaSchemaGenerator.generateSyntaxTree();
+    }
+
```
NipunaRanasinghe, line 176: Don't we have a negative test for the newly added warning?

lnash94: yeah, improve the tests for capturing the warning message.

NipunaRanasinghe (approved):

PR #950, opened 2022-04-11
[BallerinaToOpenAPI]Fix enum value representation with new lang changes 

lnash94:
Previous code generation
ballerina code
```ballerina 
enum Action {
    PATCH,
    DELETE,
    PUT,
    POST,
    GET
}
```
After the lang change, received code
```
enum:
            - '"PATCH"'
            - '"DELETE"'
            - '"PUT"'
            - '"POST"'
            - '"GET"'
```   
Expected code 
```
enum:
              - PATCH
              - DELETE
              - PUT
              - POST
              - GET
```         
Currently, the enum string value gives  with `"` . Through this fix we remove `"` from value to generate OpenAPI spec

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIComponentMapper.java
+++ b/openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIComponentMapper.java
@@ -300,9 +300,9 @@ public class OpenAPIComponentMapper {
         List<ConstantSymbol> enumMembers = enumSymbol.members();
         for (ConstantSymbol enumMember : enumMembers) {
             if (enumMember.typeDescriptor().typeKind() == TypeDescKind.SINGLETON) {
-                enums.add(enumMember.typeDescriptor().signature());
+                enums.add(enumMember.typeDescriptor().signature().replaceAll("\"", ""));
             } else {
-                enums.add(enumMember.constValue().toString().trim());
+                enums.add(enumMember.constValue().toString().trim().replaceAll("\"", ""));
             }
         }
         property.setEnum(enums);
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ group=io.ballerina
 version=1.0.3-SNAPSHOT
 
 #dependency
-ballerinaLangVersion=2201.0.3-rc1.5
+ballerinaLangVersion=2201.0.2-20220409-021000-c54e35cd
 testngVersion=7.4.0
 slf4jVersion=1.7.30
 org.gradle.jvmargs=-Xmx4096M


openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIComponentMapper.java
```diff
@@ -300,9 +300,9 @@ private Schema mapEnumValues(EnumSymbol enumSymbol) {
         List<ConstantSymbol> enumMembers = enumSymbol.members();
         for (ConstantSymbol enumMember : enumMembers) {
             if (enumMember.typeDescriptor().typeKind() == TypeDescKind.SINGLETON) {
-                enums.add(enumMember.typeDescriptor().signature());
+                enums.add(enumMember.typeDescriptor().signature().replaceAll("\"", ""));
```
MaryamZi, line 303: But this is a breaking change, right? I think we need to check the reason for this change first and fix here only if intentional (e.g., a bug fix). Could be a semantic API change or a lang change.

cc @gimantha @dulajdilshan

lnash94: @MaryamZi yeah , when I'm using `2201.0.3-rc1.4` , it returns `enumMember.typeDescriptor().signature()` , `PATCH`. 
But with new version it returns  `"PATCH"`

dulajdilshan: Here the behaviour of the Semantic API is correct. The signature of the string literals must contain `"`. I think adding this change on Open API side  is okay since this doesn't interfere with the Semantic API behaviour

dulajdilshan: The reason for this change is that previously we missed `"` in string literals. We added this with this PR https://github.com/ballerina-platform/ballerina-lang/pull/35640

MaryamZi: But this is the signature, right? Not the value?

If we had 

```ballerina
const FOO = "foo";
```

if we get the signature of type `FOO`, wouldn't/shouldn't it be `FOO`?

In https://github.com/ballerina-platform/ballerina-lang/issues/35407, there was no type definition/constant declaration for `"parent"`, so there we have to use `"parent"`. This is different, right?

dulajdilshan: Yeah. Its `"FOO". Here, shouldn't the typedef of the enum contain `"` since it is a string value?

MaryamZi:
```ballerina
enum Status {
    OPEN,
    CLOSED
}
```

is the same as

```ballerina
const OPEN = "OPEN";
const CLOSED = "CLOSED";

type Status OPEN|CLOSED;
```

When we do `io.ballerina.compiler.api.symbols.EnumSymbol#members`, shouldn't the behaviour be the same as that of union member types?

lnash94:
Currently, reverted the main PR in ballerina lang that cause to have this issue. Therefore we can close this PR in the favour of this [PR](https://github.com/ballerina-platform/ballerina-lang/pull/35798).  Please reopen if the reverted changes were not working properly.

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIComponentMapper.java
dulajdilshan: @lnash94 Can you try using `enumMember.get(0).signature()` this instead of  `enumMember.typeDescriptor().signature()`

lnash94: Tried with `enumMember.signature()`,  it will return Enum constant instead of its value,  when we have an enum with a value like below way,
Enum with value:
```ballerina
enum Action {
    GET = "get" ,
    POST = "post",
    PUT = "put",
    DELETE = "delete",
    PATCH = "patch"
}
```
expected:
```openapi
enum:
            - patch
            - delete
            - put
            - post
            - get
```
Recived :
```openapi
enum:
          - PATCH
          - DELETE
          - PUT
          - POST
          - GET
```

PR #959, opened 2022-04-29
Add double quote as value in record default value

kapilraajP:
## Purpose
- Fix https://github.com/ballerina-platform/openapi-tools/issues/958

**Default value of `"` in Schemas** 
```yaml
components:
  schemas:
    Csv:
      properties:
        textQualifier:
          type: string
          default: ' " '
```
**Generated code** 
```bal
public type Csv record {
    string textQualifier? = "\"";
};
```
## Security checks
 - Ran FindSecurityBugs plugin and verified report? yes


openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/TypeGeneratorUtils.java
```diff
@@ -163,8 +163,13 @@ public static List<Node> addRecordFields(List<String> required, Set<Map.Entry<St
                     if (field.getValue().getDefault() != null) {
                         Token defaultValue;
                         if ((field.getValue().getType()).equals(STRING)) {
-                            defaultValue = AbstractNodeFactory.createIdentifierToken("\"" +
-                                    field.getValue().getDefault().toString() + "\"");
+                            if (field.getValue().getDefault().toString().equals("\"")) {
```
lnash94, line 166: Don't we need to trim this default value, if in case we have something like this with extra space ` default: ' "  '` ?

aneeshafedo (approved):

openapi-cli/src/main/java/io/ballerina/openapi/generators/schema/TypeGeneratorUtils.java
kapilraajP: Trim the white spaces

lnash94 (approved):

Issue #849, opened 2022-01-12
Revamp remote function generation for operations with wild card (*/*) type request bodies


aneeshafedo:
**Description:**

Sample OpenAPI
```yaml
  /pets:
    post:
      summary: Create a pet
      operationId: createPet
      requestBody:
        content:
          '*/*':
            example:
              summary: Any type of entity body
      responses:
        '201':
          description: Null response
```

Generated remote function
```bal
    remote isolated function createPet(byte[] payload) returns http:Response|error {
        string resourcePath = string `/pets`;
        http:Request request = new;
        request.setPayload(payload, "*/*");
        http:Response response = check self.clientEp->post(resourcePath, request);
        return response;
    }
```

Expected remote function
```bal
    remote isolated function createPet(http:Request request) returns http:Response|error {
        string resourcePath = string `/pets`;
        http:Response response = check self.clientEp->post(resourcePath, request);
        return response;
    }
```
**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

aneeshafedo:
In the current code generation not only when request type is `*/*`, even when the response content type is `*/*`, the return type is generated `byte[]`. What would be the suitable code generation in such scenario. Please refer to the below sample 

**Sample OpenAPI** 

```yaml
  /pets/my:
    post:
      summary: Create a pet
      operationId: createMyPet
      tags:
        - pets
      requestBody:
        description: "Request to add a pet"
        content:
          '*/*':
            example:
              summary: Any type of entity body
      responses:
        '200':
            description: Successful operation
            content:
              '*/*':
                example:
                  summary: My Test
```

**Generated code (bal 2201.3.0)**

```bal
    remote isolated function createMyPet(byte[] payload) returns byte[]|error {
        string resourcePath = string `/pets/my`;
        http:Request request = new;
        request.setPayload(payload, "*/*");
        byte[] response = check self.clientEp->post(resourcePath, request);
        return response;
    }
```

**Proposed Approach 1** 
Change the return type from `byte[]` to `http:Response`

```bal
    remote isolated function createMyPet(http:Request request) returns http:Response|error {
        string resourcePath = string `/pets/my`;
        http:Response response = check self.clientEp->post(resourcePath, request);
        return response;
    }
```

NipunaRanasinghe:
[...]

Issue #636, opened 2021-10-06
Create record according to OAS `additionalProperties`  when it available


lnash94:
**Description:**
Currently, openAPI tool generated Open records for all the given schemas. OAS has a schema field called `additionalProperty`
this field uses to make ware that schema can be allowed to have any field except given properties.
Using that field we can add which type of  record we need to be created for schema

- [ ] When `additionalProperty = true` record should be open record.
```openapi
type: object
properties:
  name:
    type: string
required:
  - name
additionalProperties: true
 
````
or 
```openapi
type: object
properties:
  name:
    type: string
required:
  - name
additionalProperties {}
```
```ballerina
type MySchema record {
    string name;
};
```
- [ ] When `additionalProperties` has special field. 
```openapi
type: object
properties:
  name:
    type: string
required:
  - name
additionalProperties:
  type: string
```
```ballerina
type MySchema record {|
    string name;
    string ...
|};
```
- [ ] when `additionalProperties = false` , record map to open

```openapi
type: object
properties:
  name:
    type: string
required:
  - name
additionalProperties: false
```
```ballerina
type MySchema record {|
    string name;
|};
```
**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

aneeshafedo:
@lnash94 What will be the openapi schema to Ballerina record mapping when `additionalProperties` section is not specified in the schema.

lnash94:
Check the OAS spec with the default behaviour of additionalProperties.

lnash94:
> @lnash94 What will be the openapi schema to Ballerina record mapping when `additionalProperties` section is not specified in the schema ?

After revisiting the behaviour, this change will change the current open record to a close record if it is not specially mentioned about `additionalProperties` details because OpenAPI/JSON Schema uses the properties keyword to validate name-value pairs that are known in advance; and uses `additionalProperties`  to validate properties that are not known?

@indikasampath2000 , @abeykoon, @LakshanSS  Cloud you please let us how this change will affect the connector process?

abeykoon:
@lnash94 I went through above description. As I understood what we have now is an open record and looking at the property we will make it a close record. 

Connector records are bound to the inputs and outputs of the operations. I do not think some customer would extend them even if they are open right now, because they know `it is regarding inputs and outputs`. 

Thus the change should be safe in my POV.

[...]

PR #1120, opened 2022-11-28
Fix function generation for wildcard (*/*) type request bodies

aneeshafedo:
Fixes https://github.com/ballerina-platform/openapi-tools/issues/849

## User stories
**OpenAPI schema**
```yaml
post:
      summary: Create a pet
      operationId: createPet
      tags:
        - pets
      requestBody:
        description: "Request to add a pet"
        content:
          '*/*':
            schema:
              $ref: "#/components/schemas/Pets"
      responses:
        '201':
          description: Null response
```

**Generated remote function**

```bal
    # Create a pet
    #
    # + request - Request to add a pet 
    # + return - Null response 
    remote isolated function createPet(http:Request request) returns http:Response|error {
        string resourcePath = string `/pets`;
        // TODO: Update the request as needed;
        http:Response response = check self.clientEp->post(resourcePath, request);
        return response;
    }
```

## Security checks


openapi-core/src/main/java/io/ballerina/openapi/core/generators/client/FunctionBodyGenerator.java
```diff
@@ -645,10 +646,12 @@ private void createRequestBodyStatements(boolean isHeader, List<StatementNode> s
 
         //Create Request statement
         Map.Entry<String, MediaType> next = iterator.next();
-        VariableDeclarationNode requestVariable = GeneratorUtils.getSimpleStatement("http:Request",
-                "request", "new");
-        statementsList.add(requestVariable);
-        if (next.getValue() != null) {
+        if (!next.getKey().contains(ANY_TYPE)) {
+            VariableDeclarationNode requestVariable = GeneratorUtils.getSimpleStatement("http:Request",
+                    "request", "new");
```
NipunaRanasinghe, line 651: Is it possible to reuse the defined constants for `http:Request` and `request` in here?

openapi-core/src/main/java/io/ballerina/openapi/core/GeneratorUtils.java
```diff
@@ -426,7 +427,32 @@ public static String getBallerinaMediaType(String mediaType) {
...
+            case MediaType.TEXT_HTML:
+            case MediaType.TEXT_PLAIN:
+                return STRING_KEYWORD.stringValue();
+            case IMAGE_PNG:
+            case MediaType.APPLICATION_OCTET_STREAM:
+            case APPLICATION_PDF:
+            case ANY_TYPE:
+                return HTTP_REQUEST;
             default:
                 return SyntaxKind.JSON_KEYWORD.stringValue();
```
lnash94, line 457: Can't we map all the unhandled payload content types to `http:Request` by setting the default behaviour as `http:Request` instead of `json` ?

aneeshafedo: Was there a special requirement to make the default is `json` ?

aneeshafedo: @NipunaRanasinghe

lnash94: Normally we used `http:Request` where we can not handle and predict the request payload type. Therefore making `http:Request` default in the switch case makes sense for all the wild card behaviours as well :) WDYT?

openapi-core/src/main/java/io/ballerina/openapi/core/GeneratorUtils.java
```diff
@@ -409,9 +410,9 @@ public static OpenAPI getOpenAPIFromOpenAPIV3Parser(Path definitionPath) throws
     }
 
     /**
-     * Generate BallerinaMediaType for all the mediaTypes.
+     * Generate BallerinaMediaType for all the return mediaTypes.
      */
-    public static String getBallerinaMediaType(String mediaType) {
+    public static String getBallerinaMediaTypeForReturn(String mediaType) {
```
lnash94, line 415: Shall we implement one common util with return type `Optional<String>`, instead of repeating same content in these two functions `getBallerinaMediaTypeForReturn(String mediaType))`, `getBallerinaMediaTypeForRequest(String mediaType)` because the return and request have same media type support from ballerina when going to bind data.
 

- We can return an empty optional type if in case we encounter any unhandled scenario (most of the case default behaviour) then we can do an optional check where this util uses and do the rest of the string allocation according to the result.
ex: when util use in return type generator
`string returnType = getBallerinaMediaType(type).isPresent() ?  getBallerinaMediaType(type).get() : HTTP_RESPONSE `

openapi-core/src/main/java/io/ballerina/openapi/core/generators/client/FunctionBodyGenerator.java
aneeshafedo: Yes. Updated

openapi-core/src/main/java/io/ballerina/openapi/core/GeneratorUtils.java
aneeshafedo: This method `getBallerinaMediaType` has 13 usages and in each place doing the check for `wildcard` media types seems redundant. Therefore added a flag to identify whether the mapping the for a request or for a response. Please review.

lnash94: +1

openapi-core/src/main/java/io/ballerina/openapi/core/GeneratorUtils.java
aneeshafedo: Added a new issue to address this comment. https://github.com/ballerina-platform/openapi-tools/issues/636

NipunaRanasinghe (approved):
LGTM

NipunaRanasinghe:
@lnash94 what are the other branches that we should send duplicate PRs of this?

lnash94:
> @lnash94 what are the other branches that we should send duplicate PRs of this?

We need to send the fix to the 1.3.x  branch.  If we continue providing patch releases for 2201.2.x from ballerina lang, then we need to provide a fix for 1.2.x branch too

PR #1156, opened 2023-01-17
[1.3.x] Fix for 201 status code mapping in service resource function

lnash94:
Fix #https://github.com/ballerina-platform/openapi-tools/issues/1136

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/openapi-cli/src/test/java/io/ballerina/openapi/generators/service/ReturnTypeTests.java
+++ b/openapi-cli/src/test/java/io/ballerina/openapi/generators/service/ReturnTypeTests.java
@@ -167,7 +167,6 @@ public class ReturnTypeTests {
         OpenAPI openAPI = GeneratorUtils.getOpenAPIFromOpenAPIV3Parser(definitionPath);
         BallerinaServiceGenerator ballerinaServiceGenerator = new BallerinaServiceGenerator(openAPI, filter);
         syntaxTree = ballerinaServiceGenerator.generateSyntaxTree();
-
         CommonTestFunctions.compareGeneratedSyntaxTreewithExpectedSyntaxTree("response/scenario_14_rs.bal", syntaxTree);
     }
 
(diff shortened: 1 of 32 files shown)


openapi-cli/src/test/java/io/ballerina/openapi/generators/service/ReturnTypeTests.java
```diff
@@ -283,4 +282,13 @@ public void singleResponseWithOutContent() throws IOException, BallerinaOpenApiE
         syntaxTree = ballerinaServiceGenerator.generateSyntaxTree();
         CommonTestFunctions.compareGeneratedSyntaxTreewithExpectedSyntaxTree("response/content_null.bal", syntaxTree);
     }
+
+    @Test
```
NipunaRanasinghe, line 286: Better to add a description for the test case as the method name seems not self-explanatory.

openapi-cli/src/test/resources/generators/service/ballerina/response/scenario_14_rs.bal
```diff
@@ -3,6 +3,6 @@ import ballerina/http;
 listener http:Listener ep0 = new (80, config = {host: "petstore.openapi.io"});
 
 service /v1 on ep0 {
-    resource function post pets() returns User|BadRequestError|NotFoundError {
+    resource function post pets() returns OkUser|BadRequestError|NotFoundError {
```
NipunaRanasinghe, line 6: Don't we assert whether the `OkUser` record is having the expected structure?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ReturnTypeGenerator.java
```diff
@@ -287,7 +291,11 @@ private TypeDescriptorNode handleMultipleResponse(ApiResponses responses)
                 TypeDescriptorNode record;
                 record = returnNode.orElseGet(
                         () -> createSimpleNameReferenceNode(createIdentifierToken(ANYDATA)));
-                if (responseCode.equals(GeneratorConstants.HTTP_200)) {
+                //Check the default behaviour for return type according to POST method.
+                boolean isWithOutStatusCode =
+                        (httpMethod.equals(POST) && responseCode.equals(GeneratorConstants.HTTP_201)) ||
+                        (!httpMethod.equals(POST) && responseCode.equals(GeneratorConstants.HTTP_200));
```
NipunaRanasinghe, line 297: Shouldn't we have the L297 condition as below or, am I missing something?
```suggestion
                        (httpMethod.equals(POST) && responseCode.equals(GeneratorConstants.HTTP_201)) ||
                        (httpMethod.equals(OK) && responseCode.equals(GeneratorConstants.HTTP_200));
```

lnash94: This condition is for checking the two conditions that the return type can only be with its schema type without its status code.
1. when the POST method has status code 201 (Accepted)
2. when the other methods have status code 200(Ok) (line 297)
Therefore Line297 we need to check whether it is not POST and with 200. the suggested condition we are unable to use it because it checks the Ok that is already a 200 status code but needs to replace by checking the http method. 🙂

NipunaRanasinghe (approved):
LGTM

PR #1177, opened 2023-02-01
[master] Support OpenAPI operations with empty response in service generation

aneeshafedo:
Resolves https://github.com/ballerina-platform/openapi-tools/issues/1174

## Goals

**Sample OpenAPI**
```yaml
  /pets:
    get:
      summary: List all pets
      description: Show a list of pets in the system
      operationId: listPets
      tags:
        - pets
        - list
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: integer
            format: int32
      responses: {}
```

**Generated resource in service**

```bal
    resource function get pets(int? 'limit) returns http:Response {
    }
```

### DIFF
--- a/openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ReturnTypeGenerator.java
+++ b/openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ReturnTypeGenerator.java
@@ -133,11 +133,14 @@ public class ReturnTypeGenerator {
                 //handle multiple response scenarios ex: status code 200, 400, 500
                 TypeDescriptorNode type = handleMultipleResponse(responses);
                 returnNode = createReturnTypeDescriptorNode(createToken(RETURNS_KEYWORD), annotations, type);
-            } else {
+            } else if (responses.size() == 1) {
                 //handle single response
                 Iterator<Map.Entry<String, ApiResponse>> responseIterator = responses.entrySet().iterator();
                 Map.Entry<String, ApiResponse> response = responseIterator.next();
                 returnNode = handleSingleResponse(annotations, response);
+            } else {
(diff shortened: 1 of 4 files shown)


openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ReturnTypeGenerator.java
```diff
@@ -133,11 +133,14 @@ public ReturnTypeDescriptorNode getReturnTypeDescriptorNode(Map.Entry<PathItem.H
                 //handle multiple response scenarios ex: status code 200, 400, 500
                 TypeDescriptorNode type = handleMultipleResponse(responses);
                 returnNode = createReturnTypeDescriptorNode(createToken(RETURNS_KEYWORD), annotations, type);
-            } else {
+            } else if (responses.size() == 1) {
```
lnash94, line 136: Can a response have `null` as a value?

NipunaRanasinghe (approved):

openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ReturnTypeGenerator.java
aneeshafedo: An endpoint in a OpenAPI contract needs to have the `responses:` sections. Otherwise it gives below error from the OpenAPI parser 

```
io.ballerina.openapi.core.exception.BallerinaOpenApiException: OpenAPI definition has errors: 
attribute paths.'/pets'(get).responses is missing
```

But as given in this issue there can be occasions where the `responses:` section has defined but its empty. (`responses : {}`). In such cases in the we get an empty map for responses.

Ex: `ApiResponses responses = operation.getValue().getResponses();`

here response is a `LinkedHashMap` with size zero. 

Could not find a occasion where we get `responses == null`

lnash94 (approved):

PR #1223, opened 2023-03-14
[openapi-fixes-1.4.x] Fix for default value assign in boolean schema

lnash94:
```openapi
        type:
          type: string
        isTrue:
          type: boolean
          default: true
```
ballerina
```ballerina
    string 'type?;
    boolean isTrue = true;
```
Note: Still remaining to support default value mapping structured type (ex: Record) in openapi to ballerina .

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
+++ b/openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
@@ -49,6 +49,7 @@ import io.ballerina.openapi.core.generators.schema.ballerinatypegenerators.Union
 import io.ballerina.openapi.core.generators.schema.model.GeneratorMetaData;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.media.ArraySchema;
+import io.swagger.v3.oas.models.media.BooleanSchema;
 import io.swagger.v3.oas.models.media.ComposedSchema;
 import io.swagger.v3.oas.models.media.IntegerSchema;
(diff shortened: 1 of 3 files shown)


openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
```diff
@@ -255,8 +256,9 @@ private static RecordFieldWithDefaultValueNode getRecordFieldWithDefaultValueNod
                 defaultValueToken = AbstractNodeFactory.createIdentifierToken("\"" +
                         fieldSchema.getDefault().toString() + "\"");
             }
-        } else if (!defaultValue.matches("^[0-9]*$") &&
-                !(defaultValue.startsWith("[") && defaultValue.endsWith("]"))) {
+        } else if (!defaultValue.matches("^[0-9]*$") && !defaultValue.matches("^(\\d*\\.)?\\d+$")
```
NipunaRanasinghe, line 259: can't simplify the double negatives in here? 
```suggestion
        } else if (defaultValue.matches("[0-9]*$") && defaultValue.matches("(\\d*\\.)?\\d+$")
```

lnash94: We can't simplify the double negative here since one is for capturing integer numbers and the other for capturing decimals and float. Therefore given input should be fulfilled these two regexes to become strings. If it is the string we suppose to add `"` to the default value since swagger didn't provide value within `"`otherwise we give the value as it is.

PR #1371, opened 2023-05-27
[master] Add fix for removing unnecessary comment from generated client code

### LINKED ISSUE #1362: [Draft]Removing unnecessary comment generation for client code generation
**Description:**
If the user-provided specification is missing details in the **requestBody** for the POST method. It is generated to prompt the user to update the **requestBody** details. Generally, for the POST method, a request body is expected in ballerina. 

```openapi
paths:
  /foo:
    post:
      operationId: postFoo
      responses:
        "500":
          description: Internal server error
          content:
            text/plain:
              schema:
                type: string
```
Generated Ballerina code :
```
resource isolated function post foo() returns http:Response|error {
        string resourcePath = string `/foo`;
        http:Request request = new;
        //TODO: Update the request as needed;
        http:Response response = check self.clientEp-> post(resourcePath);
        return response;
    }
```
Since tool-generated client code is not supposed to modify by the user, this comment makes room for user confusion and we can not expect the user to go through the code and modify it. 
To overcome this we have two options

1. Provide a warning instead of generating this comment inside the function body 
  -  But still user can write their openAPI specification without request body for POSTs since requiring a body comes not from [RFC 7320](https://datatracker.ietf.org/doc/html/rfc7230) but from [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.3), where the POST method is defined.

2. We can ignore generating a warning or a comment.

lnash94:
Fix #1362

### DIFF
--- a/openapi-cli/src/test/resources/expected_gen/client_filtered_by_tags.bal
+++ b/openapi-cli/src/test/resources/expected_gen/client_filtered_by_tags.bal
@@ -51,7 +51,6 @@ public isolated client class Client {
     remote isolated function createPet() returns http:Response|error {
         string resourcePath = string `/pets`;
         http:Request request = new;
-        //TODO: Update the request as needed;
         http:Response response = check self.clientEp-> post(resourcePath, request);
         return response;
     }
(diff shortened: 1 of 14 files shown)


NipunaRanasinghe (approved):
LGTM

Issue #1384, opened 2023-06-02
Provide Record type support for Query parameter in Service generation 


lnash94:
**Description:**
Supporting Record type as a query parameter will be available with the Http module along with the Ballerina Swan Lake 2201.6.0 version. Therefore there is the possibility to map Object schema into a record when it has a Query parameter scheme. 

- When the query parameter has an inline object
- When the query parameter has a reference object
- When the query parameter has an object array

**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

NipunaRanasinghe:
Closing with https://github.com/ballerina-platform/openapi-tools/pull/1389 and https://github.com/ballerina-platform/openapi-tools/pull/1393

PR #1389, opened 2023-06-12
[1.5.x]Add query object support for the service generation

lnash94:
Fix https://github.com/ballerina-platform/openapi-tools/issues/1384

Through this fix we generate an open record for some scenarios, this open record support will be received via this https://github.com/ballerina-platform/ballerina-standard-library/issues/4541

### DIFF
--- a/openapi-cli/src/test/resources/generators/service/ballerina/query/object_query.bal
+++ b/openapi-cli/src/test/resources/generators/service/ballerina/query/object_query.bal
@@ -1,0 +1,50 @@
+import ballerina/http;
+
+listener http:Listener ep0 = new (443, config = {host: "virtserver.swaggerhub.com"});
+
+service /lnash94/QueryParam/'1\.0\.0 on ep0 {
+    # searches inventory
+    #
+    # + required_query - 1. required query param with inline object
+    # + optional_query - 2. optional query param with inline object
+    # + default_query - 3. default query param with inline object
+    # + reference_query - 4. reference query param
+    # + required_nullable - 5. required query param with inline object and nullable true
+    # + optional_nullable - 6. optional query param with inline object
+    # + default_nullable - 7. default query param with inline object
+    # + reference_nullable - 8. reference_nullable query param won't consider nullable details when it has reference in swagger
+    # + return - returns can be any of following types
+    # http:Ok (search results matching criteria)
+    # http:BadRequest (bad input parameter)
+    resource function get inventory(Required_query required_query, Optional_query? optional_query, Manufacturer? reference_query, Required_nullable? required_nullable, Optional_nullable? optional_nullable, Manufacturer? reference_nullable, Default_query default_query = {"limit":"9","id":9}, Default_nullable default_nullable = {"limit":"9","id":9}) returns http:Ok|http:BadRequest {
+    }
+    # Update inventory
+    #
+    # + primitive - 14. object query parameter with additional property with primitive value
+    # + return - Created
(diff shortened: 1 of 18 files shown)


NipunaRanasinghe (approved):

openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ServiceDiagnosticMessages.java
```diff
@@ -33,7 +33,9 @@ public enum ServiceDiagnosticMessages {
             "Query parameters with no array item type can not be mapped to Ballerina resource" +
                     " query parameters.", DiagnosticSeverity.ERROR),
     OAS_SERVICE_102("OAS_SERVICE_102",
-            "Query parameters with type '%s' can not be mapped to the Ballerina query parameters.",
+            "Type '%s' is not a valid query parameter type in Ballerina. " +
+                    "The supported types are string, int, float, boolean, decimal, array types of the aforementioned " +
+                    "types and map<json>.",
```
aneeshafedo, line 38: Are we planning to update this message including records when the open record fix is given in update 6 ?

openapi-cli/src/test/resources/generators/service/ballerina/query/object_query.bal
```diff
@@ -0,0 +1,50 @@
...
+    # + default_query - 3. default query param with inline object
+    # + reference_query - 4. reference query param
+    # + required_nullable - 5. required query param with inline object and nullable true
+    # + optional_nullable - 6. optional query param with inline object
+    # + default_nullable - 7. default query param with inline object
+    # + reference_nullable - 8. reference_nullable query param won't consider nullable details when it has reference in swagger
+    # + return - returns can be any of following types
+    # http:Ok (search results matching criteria)
+    # http:BadRequest (bad input parameter)
+    resource function get inventory(Required_query required_query, Optional_query? optional_query, Manufacturer? reference_query, Required_nullable? required_nullable, Optional_nullable? optional_nullable, Manufacturer? reference_nullable, Default_query default_query = {"limit":"9","id":9}, Default_nullable default_nullable = {"limit":"9","id":9}) returns http:Ok|http:BadRequest {
```
aneeshafedo, line 19: `Default_nullable default_nullable = {"limit":"9","id":9}` Will giving default record work in a scenario where there is a nested record ?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ServiceDiagnosticMessages.java
lnash94: As discussed, this implementation won't be excluding the code generation for open records. Ideally, it should be compatible with the HTTP module version including the [fix for open records](https://github.com/ballerina-platform/ballerina-standard-library/issues/4541).

aneeshafedo: I meant that in this message it says that only these types `string, int, float, boolean, decimal` are supported. When are we going to include `record` in that ?

aneeshafedo (approved):

openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/ServiceDiagnosticMessages.java
lnash94: Yeah, this error message aligned with the http module's existing message and I checked with them whether we need to modify it with `record`, and they informed that `map<json>` indicates the `record` behaviour so this message won't be modified.

PR #1508, opened 2023-08-18
[master] Fix for avoiding invalid constraint validation for `byte[]` OpenAPI to Ballerina client generation

lnash94:
## Automation tests
 - Unit tests - Added
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
+++ b/openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
@@ -80,7 +80,14 @@ import static io.ballerina.compiler.syntax.tree.SyntaxKind.QUESTION_MARK_TOKEN;
 import static io.ballerina.compiler.syntax.tree.SyntaxKind.SEMICOLON_TOKEN;
 import static io.ballerina.openapi.core.GeneratorConstants.BALLERINA;
 import static io.ballerina.openapi.core.GeneratorConstants.CONSTRAINT;
+import static io.ballerina.openapi.core.GeneratorConstants.DECIMAL;
+import static io.ballerina.openapi.core.GeneratorConstants.FLOAT;
+import static io.ballerina.openapi.core.GeneratorConstants.INT;
+import static io.ballerina.openapi.core.GeneratorConstants.INT_SIGNED32;
 import static io.ballerina.openapi.core.GeneratorConstants.NULL;
+import static io.ballerina.openapi.core.GeneratorConstants.STRING;
+import static io.ballerina.openapi.core.GeneratorUtils.convertOpenAPITypeToBallerina;
+import static io.ballerina.openapi.core.GeneratorUtils.getOpenAPIType;
 
 /**
  * Contains util functions needed for schema generation.
@@ -163,7 +170,7 @@ public class TypeGeneratorUtils {
     public static ImmutablePair<List<Node>, Set<String>> updateRecordFieldListWithImports(
             List<String> required, List<Node> recordFieldList, Map.Entry<String, Schema<?>> field,
             Schema<?> fieldSchema, NodeList<Node> schemaDocNodes, IdentifierToken fieldName,
-            TypeDescriptorNode fieldTypeName) {
+            TypeDescriptorNode fieldTypeName) throws BallerinaOpenApiException {
 
         return updateRecordFieldListWithImports(required, recordFieldList, field, fieldSchema, schemaDocNodes,
                 fieldName,
@@ -173,7 +180,7 @@ public class TypeGeneratorUtils {
     public static ImmutablePair<List<Node>, Set<String>> updateRecordFieldListWithImports(
             List<String> required, List<Node> recordFieldList, Map.Entry<String, Schema<?>> field,
             Schema<?> fieldSchema, NodeList<Node> schemaDocNodes, IdentifierToken fieldName,
-            TypeDescriptorNode fieldTypeName, PrintStream outStream) {
+            TypeDescriptorNode fieldTypeName, PrintStream outStream) throws BallerinaOpenApiException {
 
         MarkdownDocumentationNode documentationNode = createMarkdownDocumentationNode(schemaDocNodes);
         Set<String> imports = new HashSet<>();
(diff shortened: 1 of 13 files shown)


openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
```diff
@@ -276,12 +283,20 @@ private static RecordFieldWithDefaultValueNode getRecordFieldWithDefaultValueNod
...
      */
-    public static AnnotationNode generateConstraintNode(String typeName, Schema<?> fieldSchema) {
+    public static AnnotationNode generateConstraintNode(String typeName, Schema<?> fieldSchema)
+            throws BallerinaOpenApiException {
         if (isConstraintAllowed(typeName, fieldSchema)) {
-            if (GeneratorUtils.isStringSchema(fieldSchema)) {
+            String ballerinaType = convertOpenAPITypeToBallerina(fieldSchema);
+            //The current tool still supports byte and binary formats in string types.
+            //The rest of the string formats are mapped to the string type. Therefore, we do not provide constraint
+            //support for byte and binary formats.
```
NipunaRanasinghe, line 292:
```suggestion
            // For openAPI field schemas having 'string' type, constraints generation will be skipped when the counterpart Ballerina type is non-string (e.g. for string schemas with format 'binary' or 'byte', the counterpart ballerina type is 'byte[]', hence any string constraints cannot be applied)
```

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
```diff
@@ -276,12 +283,20 @@ private static RecordFieldWithDefaultValueNode getRecordFieldWithDefaultValueNod
...
+            //TODO: Please verify if constraints are supported when attempting to add new string format type support.
```
NipunaRanasinghe, line 293: Do we need this todo? :) IINM such cases will be automatically handled as now we validate the counterpart ballerina type as well, before adding the annotation. WDYT?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
```diff
@@ -276,12 +283,20 @@ private static RecordFieldWithDefaultValueNode getRecordFieldWithDefaultValueNod
...
+            if ((GeneratorUtils.isStringSchema(fieldSchema) && ballerinaType.equals(STRING))) {
```
NipunaRanasinghe, line 294: Do we still need to check if its a string schema? it seems redundant to me. @aneeshafedo WDYT?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
```diff
@@ -276,12 +283,20 @@ private static RecordFieldWithDefaultValueNode getRecordFieldWithDefaultValueNod
...
                 // Attributes : maxLength, minLength
                 return generateStringConstraint(fieldSchema);
-            } else if (GeneratorUtils.isNumberSchema(fieldSchema) || GeneratorUtils.isIntegerSchema(fieldSchema)) {
+            } else if (GeneratorUtils.isNumberSchema(fieldSchema) && (ballerinaType.equals(DECIMAL) ||
+                    ballerinaType.equals(FLOAT)) || (GeneratorUtils.isIntegerSchema(fieldSchema) &&
+                    (ballerinaType.equals(INT) || ballerinaType.equals(INT_SIGNED32)))) {
```
NipunaRanasinghe, line 299: same as above

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
aneeshafedo: No we don't. `ballerinaType.equals(STRING)` would be enough.

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
aneeshafedo: Yes. With this new approach we do not need to worry about the OpenAPI formats since we generate constraints based on the type of the field anyway. If that type is not supported in constraints then that will not be generated.

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
lnash94: During the call, we discussed the addition of these checks, as `isStringSchema` only verifies the type of the schema within it and doesn't capture the format details. Let me check if we can eliminate the schema check and retain the remaining Ballerina type.

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
lnash94: We included this comment for future reference, in case someone needs to address the remaining string format types. That person may benefit from the context of the constraint. 🙂

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
lnash94: Updated here https://github.com/ballerina-platform/openapi-tools/pull/1508/commits/59db7ee7bb123c4db7d1fb4d5e4b9a83ee12c240

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
lnash94: Updated here https://github.com/ballerina-platform/openapi-tools/pull/1508/commits/59db7ee7bb123c4db7d1fb4d5e4b9a83ee12c240

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
aneeshafedo: The comment added here seems the better place to warn in such case https://github.com/ballerina-platform/openapi-tools/blob/8f88e4116a61a52b2101fada200944fa93e086dc/openapi-core/src/main/java/io/ballerina/openapi/core/GeneratorConstants.java#L317

NipunaRanasinghe: @aneeshafedo yes this is the place we should have the comment ideally

aneeshafedo (approved):
LGTM

openapi-core/src/main/java/io/ballerina/openapi/core/generators/schema/TypeGeneratorUtils.java
```diff
@@ -276,12 +283,18 @@ private static RecordFieldWithDefaultValueNode getRecordFieldWithDefaultValueNod
...
+            String ballerinaType = convertOpenAPITypeToBallerina(fieldSchema);
+            // For openAPI field schemas having 'string' type, constraints generation will be skipped when
+            // the counterpart Ballerina type is non-string (e.g. for string schemas with format 'binary' or 'byte',
+            // the counterpart ballerina type is 'byte[]', hence any string constraints cannot be applied)
+            if (ballerinaType.equals(STRING)) {
                 // Attributes : maxLength, minLength
                 return generateStringConstraint(fieldSchema);
-            } else if (GeneratorUtils.isNumberSchema(fieldSchema) || GeneratorUtils.isIntegerSchema(fieldSchema)) {
+            } else if (ballerinaType.equals(DECIMAL) || ballerinaType.equals(FLOAT) || ballerinaType.equals(INT) ||
+                    ballerinaType.equals(INT_SIGNED32)) {
```
NipunaRanasinghe, line 297: What about signed16 and signed8 integers? :)

lnash94: Swagger only offers standard [integer formats](https://swagger.io/docs/specification/data-models/data-types/#numbers:~:text=with%20double%20precision.-,integer,-%E2%80%93) in signed 32 and signed 64. If the user provides a format that is not one of these options in OpenAPI, it redirects to its main data type, 'integer' (ballerina `int`). In Ballerina, the **int** type represents signed 64, while if it's signed 32, we map it to `int:Singed32`.

NipunaRanasinghe (approved):
LGTM

PR #1550, opened 2023-10-02
[1.8.x]Update README.md and Ballerina.toml files

lnash94:
## Purpose
-  Update README.md file with Java 17 support
-  Update Ballerina.toml with graalVm support

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,12 @@ For more information on the supported operations, go to [Using the OpenAPI Tools
 
 ### Setting Up the Prerequisites
 
-1. OpenJDK 11 ([Adopt OpenJDK](https://adoptopenjdk.net/) or any other OpenJDK distribution)
+1. Download and install Java SE Development Kit (JDK) version 17 (from one of the following locations).
 
-    >**Info:** You can also use [Oracle JDK](https://www.oracle.com/java/technologies/javase-downloads.html). Set the JAVA_HOME environment variable to the pathname of the directory into which you installed JDK.
+    * [Oracle](https://www.oracle.com/java/technologies/downloads/)
+    * [OpenJDK](https://adoptium.net/)
+
+      > **Note:** Set the JAVA_HOME environment variable to the path name of the directory into which you installed JDK.
    
 2. Export GitHub Personal access token with read package permissions as follows,
    ```
--- a/module-ballerina-openapi/Ballerina.toml
+++ b/module-ballerina-openapi/Ballerina.toml
@@ -3,6 +3,9 @@ org= "ballerina"
 name= "openapi"
 version= "@toml.version@"
 
+[platform.java17]
+graalvmCompatible = true
+
 [[platform.java17.dependency]]
 path = "../openapi-validator/build/libs/openapi-validator-@project.version@.jar"
 groupId = "ballerina"


dilanSachi (approved):

module-ballerina-openapi/Ballerina.toml
```diff
@@ -8,3 +8,6 @@ path = "../openapi-validator/build/libs/openapi-validator-@project.version@.jar"
 groupId = "ballerina"
 artifactId = "openapi"
 version = "@project.version@"
+
+[platform.java17]
+graalvmCompatible = true
```
NipunaRanasinghe, line 13: How did we verify that the module is graalVM compatible? Have we configured a graalVM workflow to run module test using the native executable?

README.md
```diff
@@ -22,9 +22,12 @@ For more information on the supported operations, go to [Using the OpenAPI Tools
 
 ### Setting Up the Prerequisites
 
-1. OpenJDK 11 ([Adopt OpenJDK](https://adoptopenjdk.net/) or any other OpenJDK distribution)
+1. Download and install Java SE Development Kit (JDK) version 17 (from one of the following locations).
 
-    >**Info:** You can also use [Oracle JDK](https://www.oracle.com/java/technologies/javase-downloads.html). Set the JAVA_HOME environment variable to the pathname of the directory into which you installed JDK.
+    * [Oracle](https://www.oracle.com/java/technologies/downloads/)
```
NipunaRanasinghe, line 27: Is it okay to suggest oracle distributions in our README files (as it can be misleading for our own internal users within WSO2)? @keizer619 WDYT?

module-ballerina-openapi/Ballerina.toml
dilanSachi: Since we only have ballerina code here, we can confirm its compatible right?

dilanSachi: My mistake. We do have java dependencies in the module.

NipunaRanasinghe: yeah we do use our core modules in here. so might need to come up with atleast a basic test suite to run on graalVM  before marking as `graalVM-compatible`.

lnash94: I'm curious about adding tests to verify GraalVM compatibility, especially since we have CLI and OpenAPI validator functionality that operates during compilation. I'll look into incorporating tests to ensure compatibility with GraalVM. 
Appreciate it if anyone has any references or insights related to GraalVM compatibility testing, please feel free to share them :)

NipunaRanasinghe: Looping in @TharmiganK

README.md
keizer619: Yes Lets suggest Open JDK

module-ballerina-openapi/Ballerina.toml
TharmiganK: This is tool right? Since we are not executing the bal tools as GraalVM native images it is not required to be GraalVM compatible

NipunaRanasinghe: @TharmiganK this is not the tool. We do have this `ballerina/openapi` library module to provide some annotation related to our openAPI validation support. So I think we should consider marking it as graalvm compatible (or not).

TharmiganK: I see
Are these annotations only relevant when we call OAS generation? 

If we do not have any runtime use cases we can mark it as GraalVM compatible to avoid the warnings when building (`bal build/test --native` or `bal pack`)

TharmiganK: Btw why do we have a java dependency in the `Ballerina.toml`? https://github.com/ballerina-platform/openapi-tools/blob/321ac864fe03d5328375d20fcdfc8284cec0a319/module-ballerina-openapi/Ballerina.toml#L6

Isn't this for the compiler plugin? I saw the same jar in the `CompilerPlugin.toml` : https://github.com/ballerina-platform/openapi-tools/blob/321ac864fe03d5328375d20fcdfc8284cec0a319/module-ballerina-openapi/CompilerPlugin.toml#L6

I think just removing it from the `Ballerina.toml` should work

lnash94:
Close this PR since we checking the update of gralvm compatibility and resend the needful

PR #1552, opened 2023-10-06
[constraint_support]Sync with master 

lnash94:
Make consistent with master branch change


openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIRequestBodyMapper.java
```diff
@@ -154,62 +163,174 @@ public void handlePayloadAnnotation(RequiredParameterNode payloadNode, Map<Strin
      *     resource function post pets(@http:Payload json payload){}
      *</pre>
      */
-    private void handleSinglePayloadType(RequiredParameterNode payloadNode, Map<String, Schema> schema,
-                                         RequestBody bodyParameter, String customMediaPrefix) {
+    private void handlePayloadType(TypeDescriptorNode payloadNode, Map<String, Schema> schema,
+                                   RequestBody bodyParameter) {
+        SyntaxKind kind = payloadNode.kind();
+        String mediaTypeString = getMediaTypeForSyntaxKind(payloadNode);
+        if (mediaTypeString != null || kind == SyntaxKind.UNION_TYPE_DESC ||
```
ayeshLK, line 170: Shall we take this logic into another method ?

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/service/OpenAPIRequestBodyMapper.java
```diff
@@ -154,62 +163,174 @@ public void handlePayloadAnnotation(RequiredParameterNode payloadNode, Map<Strin
...
+                    //Warning message for unsupported request payload type in Ballerina resource.
+                    IncompatibleResourceDiagnostic error = new IncompatibleResourceDiagnostic(
+                            DiagnosticMessages.OAS_CONVERTOR_117, payloadNode.location(),
+                            payloadNode.toSourceCode().trim());
+                    diagnostics.add(error);
+                    break;
+            }
+        } else {
+            //Warning message for unsupported request payload type in Ballerina resource.
+            IncompatibleResourceDiagnostic error = new IncompatibleResourceDiagnostic(
```
ayeshLK, line 255: Rather than using the `if-else` block i think w could check whether the condition failure and do an early return from there after adding the necessary warnings to the diagnostic-context.

openapi-cli/src/test/java/io/ballerina/openapi/generators/openapi/ModuleReferenceTests.java
```diff
@@ -73,7 +73,7 @@ public void testRecordReferenceWithReadOnly() throws IOException {
         TestUtils.compareWithGeneratedFile(ballerinaFilePath, "readonly.yaml");
     }
 
-    @Test ()
+    @Test (enabled = false)
```
NipunaRanasinghe, line 76: Any reason why we are disabling these tests? Listeners defined in separate files and modules won't be supported with the new changes?

ayeshLK (approved):

openapi-cli/src/test/java/io/ballerina/openapi/generators/openapi/ModuleReferenceTests.java
lnash94: Due to ongoing implementation disable these tests since it has interminate failures. we enable these tests.

PR #1570, opened 2023-11-15
[1.7.x] Fix default value mapping in the OAS generation

TharmiganK:
Related to: https://github.com/wso2-enterprise/internal-support-ballerina/issues/505

### DIFF
--- a/openapi-bal-service/src/main/java/io/ballerina/openapi/converter/utils/ConverterCommonUtils.java
+++ b/openapi-bal-service/src/main/java/io/ballerina/openapi/converter/utils/ConverterCommonUtils.java
@@ -82,6 +82,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -527,6 +528,9 @@ public class ConverterCommonUtils {
     }
 
     public static void setDefaultValue(Schema schema, String defaultValue) {
+        if (Objects.isNull(schema) || Objects.isNull(defaultValue)) {
+            return;
+        }
         try {
             ObjectMapper mapper = new ObjectMapper();
             mapper.configure(JsonParser.Feature.ALLOW_UNQUOTED_FIELD_NAMES, true);
--- a/openapi-bal-service/src/main/java/io/ballerina/openapi/converter/utils/ConverterCommonUtils.java
+++ b/openapi-bal-service/src/main/java/io/ballerina/openapi/converter/utils/ConverterCommonUtils.java
@@ -18,6 +18,9 @@
 
 package io.ballerina.openapi.converter.utils;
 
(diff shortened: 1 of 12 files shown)


openapi-cli/src/test/resources/ballerina-to-openapi/query/query_scenario09.bal
```diff
@@ -0,0 +1,20 @@
+import ballerina/http;
```
dilanSachi, line 1: license header

TharmiganK: We do not have it in any other test files. @lnash94 should we add it to all the test files?

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/utils/ConverterCommonUtils.java
```diff
@@ -522,4 +526,17 @@ public static String unescapeIdentifier(String parameterName) {
...
+
+    public static void setDefaultValue(Schema schema, String defaultValue) {
+        if (Objects.isNull(schema) || Objects.isNull(defaultValue)) {
+            return;
+        }
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            mapper.configure(JsonParser.Feature.ALLOW_UNQUOTED_FIELD_NAMES, true);
+            schema.setDefault(mapper.readValue(defaultValue, Object.class));
+        } catch (JsonProcessingException e) {
```
lnash94, line 538: What are the possible scenarios this JsonProcesingException can happen?

TharmiganK: The exception is thrown if the the string does not match the JSON string format. (Here we have allowed the json fields without quotes to support ballerina record literals)
So there should not be a case where the literal does not align with JSON format. This try-catch has been added as a saftly mechanism.

openapi-cli/src/test/resources/ballerina-to-openapi/query/query_scenario09.bal
lnash94: FTM, we can avoid adding license headers to the test file since any other test files don't include the header. In future let's refactor the whole tests with headers.

openapi-bal-service/src/main/java/io/ballerina/openapi/converter/utils/ConverterCommonUtils.java
lnash94: What is the `defaultValue` value in such cases, when the exception occur ?

TharmiganK: It is just the ballerina literal expression, in those exceptional cases we just print that expression as string. But ideally this exception will never happen

lnash94 (approved):

PR #1599, opened 2024-01-29
[Ballerina to OAS] Interceptor Support

SachinAkash01:
Fixes: [#5986](https://github.com/ballerina-platform/ballerina-library/issues/5986)
See also: [#5956](https://github.com/ballerina-platform/ballerina-library/issues/5956)


TharmiganK:
Closing this implementation since there are some gaps to fill:
1. Resource interceptor matching based on the resource method and path. We should implement a similar approach in dispatcher to compare interceptor path with target resource path. The interceptor is only invokable if the path match.
2. The logic to find the effective return type should be revisited since it is failing for some common scenarios when there are multiple interceptors.

The new implementation based on this is available in https://github.com/ballerina-platform/openapi-tools/pull/1614

PR #1620, opened 2024-03-16
[typegen-restruct] Restructure openapi core

lnash94:
## Purpose
Move all the separate packages under the generator package due to its readability and those are code generators.
**Changes:**
 - Moved **typegenerator** to inside generator package with name `type`
 - Moved **constraintgenerator** to inside generator package with name `constraint`
 - Renamed schema to **schemaOld** and service to **serviceOld** once new revamp done for the `service` package which is currently located outsit in the generator can move to inside and remove those old packages
 - Create **common** package and move Generator Util, models, Constant and exceptions to common 
 -  Move TypeHandler to **common** package since it is using both service and client, its behaviour as an intermediator in between type generator and other service/client generator 
 - Added new **constraint** package to generators
 
Action items:
- Each  model should have their own diagnostic messages and exceptions for the error-handling purpose
- Each component should be exposed by using the interface.

Sutructure diagram: https://docs.google.com/drawings/d/1x666imFbcDB9ClmEalyiehRt0TZYM4tWsd6WWSsO8hM/edit?usp=sharing


lnash94:
Test are failing , and this will be fix eventually

PR #1644, opened 2024-04-04
Add publish task for grade to publish ballerina-openapi artifact into maven

SasinduDilshara:
Add publish task for grade to publish ballerina-openapi artifact into maven

### DIFF
--- a/ballerina-to-openapi/build.gradle
+++ b/ballerina-to-openapi/build.gradle
@@ -132,3 +132,27 @@ compileJava {
         classpath = files()
     }
 }
+
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
+            groupId project.group
+            artifactId "ballerina-to-openapi"
+            version = project.version
+            artifact jar
+        }
+    }
+
+    repositories {
+        maven {
+            name = "GitHubPackages"
+            url = uri("https://maven.pkg.github.com/ballerina-platform/openapi-tools")
+            credentials {
+                username = System.getenv("packageUser")
+                password = System.getenv("packagePAT")
+            }
+        }
+    }
+}
+
+publish.dependsOn build
--- a/ballerina-to-openapi/build.gradle
+++ b/ballerina-to-openapi/build.gradle
@@ -136,7 +136,7 @@ compileJava {
 publishing {
     publications {
         mavenJava(MavenPublication) {
-            groupId project.group
+            groupId "io.ballerina.openapi"
             artifactId "ballerina-to-openapi"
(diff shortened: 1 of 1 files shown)


ballerina-to-openapi/build.gradle
```diff
@@ -132,3 +132,27 @@ compileJava {
         classpath = files()
     }
 }
+
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
```
lnash94, line 138: will this publish to GitHub packages or Is there a reason not to publish in GitHub packages also?

SasinduDilshara:
```
publishing {
    publications {
        mavenJava(MavenPublication) {
            artifact source: createArtifactZip, extension: 'zip'
        }
    }

    repositories {
        maven {
            name = "GitHubPackages"
            url = uri("https://maven.pkg.github.com/ballerina-platform/openapi-tools")
            credentials {
                username = System.getenv("packageUser")
                password = System.getenv("packagePAT")
            }
        }
    }
}
```

This is how `module-ballerina-openapi` got published, Don't I need to do the same in here

ballerina-to-openapi/build.gradle
```diff
@@ -132,3 +132,27 @@ compileJava {
...
+            groupId project.group
```
TharmiganK, line 139: Can we change the group here? The project group is `io.ballerina`. I think it is better to publish it under `io.ballerina.openapi`
```suggestion
            groupId "io.ballerina.openapi"
```

ballerina-to-openapi/build.gradle
TharmiganK: I think this should work

TharmiganK: I think this is correct

ballerina-to-openapi/build.gradle
SasinduDilshara: `Persist tool` is publish to `io.ballerina`.
https://github.com/ballerina-platform/persist-tools/blob/a487cbc0c085e572d042a74e851d1a2e51e4da3a/persist-cli/build.gradle#L151

Don't we need to have a same group id among the tools?

TharmiganK: That is the entire tool right? This is just a part of the tool

SasinduDilshara: ok, will update that

TharmiganK (approved):
LGTM

PR #1677, opened 2024-04-16
Escape blank identifiers and add warning for service blank headers

dilanSachi:
## Purpose
$subject

Related to https://github.com/ballerina-platform/ballerina-library/issues/4879


openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/GeneratorUtils.java
```diff
@@ -396,6 +396,9 @@ public static String escapeIdentifier(String identifier) {
      * @return string with new generated name
      */
     public static String getValidName(String identifier, boolean isSchema) {
+        if (identifier.isBlank()) {
```
lnash94, line 399: shall we handle blank scenarios in particular places as well? ex:  if the header is with blank name, it can not be accepted as a header name, therefore we need to provide a user warning or some other way to handle it.

dilanSachi: Updated

openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/parameter/HeaderParameterGenerator.java
```diff
@@ -84,6 +84,9 @@ public ParameterNode generateParameterNode(Parameter parameter) throws Unsupport
         Schema<?> schema = parameter.getSchema();
         String headerType = GeneratorConstants.STRING;
         TypeDescriptorNode headerTypeName;
+        if (parameter.getName().isEmpty() || parameter.getName().isBlank()) {
```
NipunaRanasinghe, line 87:
```suggestion
        if (parameter.getName().isBlank()) {
```

openapi-core/src/main/java/io/ballerina/openapi/core/generators/service/exceptions/InvalidHeaderNameException.java
```diff
@@ -0,0 +1,28 @@
...
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package io.ballerina.openapi.core.generators.service.exceptions;
+
+import io.ballerina.openapi.core.generators.service.diagnostic.ServiceDiagnostic;
+import io.ballerina.openapi.core.generators.service.diagnostic.ServiceDiagnosticMessages;
+
+public class InvalidHeaderNameException extends BallerinaServiceException {
```
NipunaRanasinghe, line 28: ditto

PR #1683, opened 2024-04-17
Add changes from the team meeting

TharmiganK:
- Changed the file name which is used to get the client native version
- Changed the status code binding option name: `--with-status-code-binding` -> `--status-code-binding`
- Add a warning when skipping `default` response in the operation with status code binding option

### DIFF
--- a/openapi-cli/src/main/java/io/ballerina/openapi/cmd/BaseCmd.java
+++ b/openapi-cli/src/main/java/io/ballerina/openapi/cmd/BaseCmd.java
@@ -51,7 +51,7 @@ public class BaseCmd {
             " with provided type . Only \"resource\"(default) and \"remote\" options are supported.")
     public String generateClientMethods;
 
-    @CommandLine.Option(names = {"--with-status-code-binding"}, description = "Generate the client methods with " +
+    @CommandLine.Option(names = {"--status-code-binding"}, description = "Generate the client methods with " +
             "status code response binding")
     public boolean statusCodeBinding;
 }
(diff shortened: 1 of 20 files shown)


openapi-cli/src/main/java/io/ballerina/openapi/cmd/BaseCmd.java
```diff
@@ -51,7 +51,7 @@ public class BaseCmd {
             " with provided type . Only \"resource\"(default) and \"remote\" options are supported.")
     public String generateClientMethods;
 
-    @CommandLine.Option(names = {"--with-status-code-binding"}, description = "Generate the client methods with " +
+    @CommandLine.Option(names = {"--status-code-binding"}, description = "Generate the client methods with " +
```
NipunaRanasinghe, line 54: Shouldn't this be `--client-status-code-binding` as it only applies to the client generation?

TharmiganK: Yes, it is applicable only for the client generation.
Since the `status-code` implies the response and response **binding** is only for the client side, I have skipped the term `client` here.
Actually we had a discussion on similar naming issues with other options as well and we have decided to revisit and fix them in the next release since we need some time to think this through.
Looping @lnash94

lnash94: Indeed, it seems that `--client-status-code-binding` serves a purpose in aligning client generation. Following our discussion today, we've reached a consensus to retain this flag for the current release. Because we anticipate revisiting the options and commands https://github.com/ballerina-platform/ballerina-library/issues/6345. to the next release. We'll be sure to document these suggestions for our forthcoming review of the flags.

lnash94 (approved):

PR #1693, opened 2024-04-22
Fix minor issues found in RC testing

TharmiganK:
- Add a property to skip publishing client native artifacts to WSO2 nexus. With this change, we have to enable this in the `gradle.properties` file, if we want to publish the artifacts with release workflow. If we do not need to release we need to mark it as `true`
- Fix optional response header mapping in the status code response. Without this change, when we have non-required headers in status code response, it is map to `map<string|string[]>`. With this change, the optional headers will be mapped as nilable record fields
- Fix printing `Responses with 204 status code cannot have a body` warning for response which does not define a request body schema
- Removed unused `OAS_COMMON_102` from the common diagnostics
- Print the client warning/error diagnostics when bal commands are executed with openAPI tool integration

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,8 +4,8 @@ version=1.9.0-SNAPSHOT
 
 # Client Native Version
 clientNativeVersion=1.0.0-SNAPSHOT
-# Mark this as true to skip publishing the client native artifacts
-skipClientNativePublish=false
+# Mark this as false to skip publishing the client native artifacts
+clientNativePublish=true
 
 #dependency
 ballerinaLangVersion=2201.9.0-20240419-152500-bd530dd2
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,6 +4,8 @@ version=1.9.0-SNAPSHOT
 
 # Client Native Version
 clientNativeVersion=1.0.0-SNAPSHOT
+# Mark this as true to skip publishing the client native artifacts
+skipClientNativePublish=false
 
 #dependency
 ballerinaLangVersion=2201.9.0-20240419-152500-bd530dd2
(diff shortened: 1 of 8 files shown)


openapi-bal-task-plugin/src/main/java/io/ballerina/openapi/bal/tool/OpenAPICodeGeneratorTool.java
```diff
@@ -461,6 +463,21 @@ private static List<GenSrcFile> generateClientFiles(OASClientConfig oasClientCon
...
+            for (ClientDiagnostic diagnostic : clientDiagnostic) {
+                createDiagnostics(toolContext, diagnostic.getMessage(), diagnostic.getCode(),
+                        diagnostic.getDiagnosticSeverity(), location);
+            }
+
+            if (clientDiagnostic.stream().anyMatch(
+                    diagnostic -> diagnostic.getDiagnosticSeverity() == DiagnosticSeverity.ERROR)) {
+                throw new ClientException("Error occurred while generating client");
+            }
+        }
```
NipunaRanasinghe, line 479: Cant we simply do
```suggestion
            for (ClientDiagnostic diagnostic : clientDiagnostic) {
                createDiagnostics(toolContext, diagnostic.getMessage(), diagnostic.getCode(),
                        diagnostic.getDiagnosticSeverity(), location);
            }

            if (clientDiagnostic.stream().anyMatch(
                    diagnostic -> diagnostic.getDiagnosticSeverity() == DiagnosticSeverity.ERROR)) {
                throw new ClientException("Error occurred while generating client");
            }
```
?

gradle.properties
```diff
@@ -4,6 +4,8 @@ version=1.9.0-SNAPSHOT
 
 # Client Native Version
 clientNativeVersion=1.0.0-SNAPSHOT
+# Mark this as true to skip publishing the client native artifacts
+skipClientNativePublish=false
```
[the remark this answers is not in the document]
TharmiganK: Fixed via [b32c1ac](https://github.com/ballerina-platform/openapi-tools/pull/1693/commits/b32c1ac0aa57f2ec3550d3ca18e4dea1e3dc5309)

openapi-bal-task-plugin/src/main/java/io/ballerina/openapi/bal/tool/OpenAPICodeGeneratorTool.java
TharmiganK: Fixed via [b32c1ac](https://github.com/ballerina-platform/openapi-tools/pull/1693/commits/b32c1ac0aa57f2ec3550d3ca18e4dea1e3dc5309)

lnash94 (approved):

PR #1697, opened 2024-04-29
[master] Fix for removing `getValidName()` usage in record Name generation for OAS to Ballerina service/client

lnash94:
## Purpose
Currently tool uses used `getValidName` method for sanitizing the variable name, record name and parameter name which the user has provided in the yaml. Because of that, we observe several issues like changing the name, overriding records etc.  with the previous discussion during the meeting of this issue https://github.com/ballerina-platform/ballerina-library/issues/4882
we decided to provide the names the user provided as it is in the code generations. 

Fixe: https://github.com/ballerina-platform/ballerina-library/issues/5140


TharmiganK (commented):
LGTM

TharmiganK (approved):

dilanSachi (approved):

openapi-integration-tests/src/test/resources/service/return/ballerina/response_has_inline_additional_properties.bal
```diff
@@ -21,17 +26,17 @@ public type StoreInventory04Response record {|
...
+    int...;
+|};
 
 public type StoreInventory05Response record {|
     User...;
 |};
 
-public type Inline_response_200 record {|
-    string name?;
+public type inline_response_map200 record {
```
NipunaRanasinghe, line 39: Now we seem to generate types names (even for inline schemas) starting with simple letters, which is not aligned with Ballerina type naming conventions. Is this done intentionally?

TharmiganK: Yes, we have decided to generate (all) type names as it is in the openapi specification.

The type names for inline schemas are generated by the parser, and at the moment we are not special casing these. Moreover in the upcoming release, we will introduce a new option in the tool to disable this behaviour of the parser(the generated code will have inline records) - the default option may be the inline records

cc @lnash94

TharmiganK (approved):

openapi-integration-tests/src/test/resources/service/return/ballerina/response_has_inline_additional_properties.bal
lnash94: Reference to this issue : https://github.com/ballerina-platform/ballerina-library/issues/5140 , as mentioned in @TharmiganK description we discussed moving forward with this change and will improve this name-changing option via in flatten feature

PR #1715, opened 2024-05-31
Fix status code response creation without body and headers

TharmiganK:


NipunaRanasinghe (commented):
@TharmiganK I'm not seeing any changes to the test suite along with this change. Does that mean this particular scenario is not captured in the existing test suite? Is that expected?

TharmiganK:
> @TharmiganK I'm not seeing any changes to the test suite along with this change. Does that mean this particular scenario is not captured in the existing test suite? Is that expected?

Yes, this is an edge case. I will add some test cases.

lnash94 (approved):

NipunaRanasinghe (approved):

PR #1717, opened 2024-06-11
[2.0.x] Add missing service meta info for the openAPI spec generation

lnash94:
Introduced below fields for the openAPI service annotation for spec generation
```ballerina
public type ServiceInformation record {|
    ...
    string email?;
    string description?;
    string contactName?;
    string contactURL?;
    string termsOfService?;
    string licenseName?;
    string licenseURL?;
|};
```
Generated openapi spec
```openapi
openapi: 3.0.1
info:
  title: Pet Store
  description: API system description
  termsOfService: http://mock-api-doc
  contact:
    name: sumudu
    url: http://mock-api-contact
    email: sumudu@abc.com
  license:
    name: ABC
    url: http://abc.com
```


ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/Constants.java
```diff
@@ -329,6 +329,14 @@ public String toString() {
...
     public static final String VERSION = "version";
     public static final String TITLE = "title";
+    public static final String EMAIL = "email";
+    public static final String DESCRIPTION = "description";
+    public static final String CONTACT_NAME = "contactName";
+    public static final String CONTACT_URL = "contactURL";
+    public static final String LICENSE_NAME = "licenseName";
+
+    public static final String LICENSE_URL = "licenseURL";
+    public static final String TERMS_OF_SERVICE = "termsOfService";
```
ayeshLK, line 339:
```suggestion
    public static final String EMAIL = "email";
    public static final String DESCRIPTION = "description";
    public static final String CONTACT_NAME = "contactName";
    public static final String CONTACT_URL = "contactURL";
    public static final String LICENSE_NAME = "licenseName";
    public static final String LICENSE_URL = "licenseURL";
    public static final String TERMS_OF_SERVICE = "termsOfService";
    
```

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/InfoMapper.java
```diff
@@ -202,31 +211,66 @@ private static String normalizeTitle(String title) {
...
-               } else if (openAPIInfo.getTitle().isPresent() && openAPIInfo.getVersion().isPresent()) {
-                   openAPI.setInfo(new Info().version(openAPIInfo.getVersion().get()).title(normalizeTitle
-                           (openAPIInfo.getTitle().get())));
-               } else if (openAPIInfo.getVersion().isPresent()) {
-                   openAPI.setInfo(new Info().version(openAPIInfo.getVersion().get()));
-               } else if (openAPIInfo.getTitle().isPresent()) {
-                   openAPI.setInfo(new Info().title(normalizeTitle(openAPIInfo.getTitle().get())));
-               }
-           }
+        Optional<MappingConstructorExpressionNode> contentOpt = annotation.annotValue();
```
ayeshLK, line 217:
```suggestion
        Optional<MappingConstructorExpressionNode> svcInfoAnnotationValue = annotation.annotValue();
```

Wouldn't this be more meaningful in this context ?

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/InfoMapper.java
```diff
@@ -266,17 +298,18 @@ private static OpenAPIInfo updateOpenAPIInfoModel(SeparatedNodeList<MappingField
...
+                            case DESCRIPTION -> infoBuilder.description(fieldValue);
+                            case CONTACT_NAME -> infoBuilder.contactName(fieldValue);
+                            case CONTACT_URL -> infoBuilder.contactURL(fieldValue);
+                            case TERMS_OF_SERVICE -> infoBuilder.termsOfService(fieldValue);
+                            case LICENSE_NAME -> infoBuilder.licenseName(fieldValue);
+                            case LICENSE_URL -> infoBuilder.licenseURL(fieldValue);
+                            default -> {
+                            }
                         }
                     }
```
ayeshLK, line 314: I think this can have an early return so that we could remove unwanted indentation.
```java
if (fieldValue.isBlank()) {
  // return from here
}
switch (fieldName) {
    case CONTRACT -> infoBuilder.contractPath(fieldValue);
    // ...
}
```

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/model/OpenAPIInfo.java
```diff
@@ -49,13 +63,48 @@ public Optional<String> getContractPath() {
         return Optional.ofNullable(this.contractPath);
     }
 
+    public Optional<String> getDescription() {
```
ayeshLK, line 66: Are all these fields are nullable ? If not lets remove `Optional` from unwanted fields.

module-ballerina-openapi/Ballerina.toml
```diff
@@ -1,10 +1,10 @@
...
-version= "1.9.0"
+version= "@toml.version@"
 
 [[platform.java17.dependency]]
-path = "../openapi-validator/build/libs/openapi-validator-1.9.0-SNAPSHOT.jar"
+path = "../openapi-validator/build/libs/openapi-validator-@project.version@.jar"
 groupId = "ballerina"
 artifactId = "openapi"
-version = "1.9.0-SNAPSHOT"
+version = "@project.version@"
```
ayeshLK, line 10:
```suggestion
version = "@project.version@"

```

module-ballerina-openapi/annotation.bal
```diff
@@ -24,6 +24,13 @@
...
+# + description - A brief description of the API, outlining its purpose, features, and any other relevant details that help users understand what the API does and how to use it.
+# + email - The email address to contact the API provider or support.
+# + contactName - The full name of the person or organization responsible for the API.
+# + contactURL - The URL to a web page with more information about the API, the provider, or support.
+# + termOfService - The URL to the terms of service for the API.
+# + licenseName - The name of the license under which the API is provided.
+# + licenseURL - The URL to the full text of the license.
 public type ServiceInformation record {|
     string contract = "";
     string[]? tags = [];
```
ayeshLK, line 36: Just checking why we have marked `tags` as `string[]?`. Is this the correct usecase ? I think `string[]` would be enough with an empty array as the default value.

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/Constants.java
lnash94: They are newly introduced fields for the `OPENAPI_ANNOTATION`

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/model/OpenAPIInfo.java
lnash94: These fields are intended to have null during the value extraction process, therefore we keep those fields enabling the null.

module-ballerina-openapi/annotation.bal
lnash94: This was introduced in the previous design. With this design, they might allow null values, although provide a default value. We will be able to revisit and make necessary adjustments separately.

ayeshLK: I think this should be revisited properly and fix. If this is nullable filed it should be represented as `string[] tags?;` rather than using `string[]?`.

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/Constants.java
ayeshLK: Then the space in line 337 does not make any sense right ?

lnash94: It was an extra line and already fixed there https://github.com/ballerina-platform/openapi-tools/pull/1717/commits/d1287cabce88bc4f0c18f8344f282bf9881eae55

module-ballerina-openapi/annotation.bal
ayeshLK: Nil means that there is no value right. That means there can be a scenario we won't have a value for this field. In that case it would be more sensible to make it optional rather than making it nillable. I think that has more ballerina-ish touch rather than marking fields nillable.

ayeshLK: I think this was designed sometime back and that is why we have followed this approach (maybe due to some limitation which was there at that time). Usually when designing APIs we prefer optional fields over nillable fields.

lnash94: yeah, usually we provide optional fields for recent implementation, will revisit this design and do needful.

module-ballerina-openapi/Ballerina.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 org= "ballerina"
 name= "openapi"
-version= "1.9.0"
+version= "@toml.version@"
 
 [[platform.java17.dependency]]
-path = "../openapi-validator/build/libs/openapi-validator-1.9.0-SNAPSHOT.jar"
+path = "../openapi-validator/build/libs/openapi-validator-@project.version@.jar"
```
TharmiganK, line 7: Why this version changed like this?

module-ballerina-openapi/CompilerPlugin.toml
```diff
@@ -3,7 +3,7 @@ id = "openapi-tools"
 class = "io.ballerina.openapi.validator.OpenAPIValidatorPlugin"
 
 [[dependency]]
-path = "../openapi-validator/build/libs/openapi-validator-1.9.0-SNAPSHOT.jar"
+path = "../openapi-validator/build/libs/openapi-validator-2.0.1-@project.version@"
```
TharmiganK, line 6: This is wrong right: `2.0.1-@project.version@`

openapi-integration-tests/src/test/resources/ballerina_sources/project_openapi_info/service_file.bal
```diff
@@ -0,0 +1,19 @@
+import ballerina/http;
+import ballerina/openapi;
+
+@openapi:ServiceInfo {
```
TharmiganK, line 4: Is this a constant annotation, if not, is there a limitation if we use variables as the values?

module-ballerina-openapi/Ballerina.toml
lnash94: It was previously committed mistakenly I guess. I revert the commit

openapi-integration-tests/src/test/resources/ballerina_sources/project_openapi_info/service_file.bal
lnash94: No, this is not a constant one, we can make it constant. Will change with separate PR by finding the impact of other usage.

TharmiganK (approved):
LGTM

PR #1722, opened 2024-06-14
[2.0.x] Add the workaround for openapi resource annotation data binding issue 

lnash94:

### DIFF
--- a/.github/workflows/central-publish.yml
+++ b/.github/workflows/central-publish.yml
@@ -2,6 +2,15 @@ name: Publish to the Ballerina central
 
 on:
   workflow_dispatch:
+    inputs:
+      environment:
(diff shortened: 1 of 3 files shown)


lnash94:
> Shall we add a test with this new annotation? If someone removes the `readonly &` part (since it's theoretically redundant with annotations), we wouldn't detect that.

Will add some integration tests for this

anuruddhal (approved):

MaryamZi (approved):
Approving the change to the record type definition as a workaround for https://github.com/ballerina-platform/ballerina-lang/issues/42933 and based on https://github.com/ballerina-platform/openapi-tools/pull/1722#issuecomment-2172419121.

PR #1734, opened 2024-06-27
Add service contract type support in OpenAPI tool

TharmiganK:
Part of: https://github.com/ballerina-platform/ballerina-library/issues/6378


ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/ServersMapperImpl.java
```diff
@@ -266,4 +271,22 @@ private static String concatenateServerURL(String host, SeparatedNodeList<Mappin
         }
         return host;
     }
+
+    private static Server getDefaultServerWithBasePath(String serviceBasePath) {
+        String serverUrl = String.format("http://{server}:{port}%s", serviceBasePath);
```
lnash94, line 276: Does this template change according to the port? ex: if the port is 443, we ideally map it to https:

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/model/ResourceFunctionDeclaration.java
```diff
@@ -0,0 +1,75 @@
+/*
+ *  Copyright (c) 2024, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
```
lnash94, line 2:
```suggestion
 *  Copyright (c) 2024, WSO2 LLC. (http://www.wso2.org).
```

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/ServersMapperImpl.java
TharmiganK: This is added as a dummy server schema for service contract type since it is not possible to get the listener details. When we implement a service with this contract, we will replace these default values from the listener.

lnash94 (approved):

lnash94 (approved):

PR #1749, opened 2024-08-07
Update swagger parser version

TharmiganK:
## Purpose

> $Subject


openapi-cli/src/test/resources/ballerina-to-openapi/advance/openapi/multiple_service_01.yaml
```diff
@@ -27,7 +27,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorPayload'
+                $ref: "#/components/schemas/ErrorPayload"
```
NipunaRanasinghe, line 30: What's the reason for this change? What will be the behaviour with the new parser version if the contract is having single quotes for the ref schema values?

TharmiganK: This change is only for the schema generation from the OAS model object. The parser works for both single and double quotes.

NipunaRanasinghe: I see. That'll be fine then

NipunaRanasinghe (approved):
LGTM

ballerina-to-openapi/build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
         exclude group: "javax.validation", module: "validation-api"
     }
     implementation "io.swagger.core.v3:swagger-models"
+    implementation "io.swagger.parser.v3:swagger-parser-safe-url-resolver:${swaggerParserVersion}"
```
lnash94, line 41: How do we use this `parser-safe-url-resolver `, is it using the swagger parser itself or do we need to use it within our implementation separately?

TharmiganK: This is a dependency of `swagger-parser`, from this latest version. There were `NoClassDefFound` errors without this as a dependency

lnash94 (approved):

PR #1766, opened 2024-08-27
Refactor type name sanitization logic for references

TharmiganK:
Fixes: https://github.com/ballerina-platform/ballerina-library/issues/6919
Fixes: https://github.com/ballerina-platform/ballerina-library/issues/6922


openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -53,6 +52,7 @@
  */
 public class OASModifier {
     List<Diagnostic> diagnostics = new ArrayList<>();
+    private static final PrintStream outStream = System.out;
```
daneshk, line 55: We usually use `System.err`  to print all messages to the console.

TharmiganK: We have used the `System.out` to print all the diagnostics we get from the client/service generations. This includes warnings and errors.

https://github.com/ballerina-platform/openapi-tools/blob/a9ce3cf2318ba258e2a4b92949341d995e6744c0/openapi-cli/src/main/java/io/ballerina/openapi/cmd/BallerinaCodeGenerator.java#L209

There are instances where we use `System.err` to print the errors. @lnash94 is there any reason behind this decision? Otherwise we can stick to print all the **error** diagnostics using `System.err` and the **warnings** using `System.out`. WDYT?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -116,137 +116,59 @@ public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi, Map<String, Strin
...
-                                                     String modifiedName, Map<String, Schema> modifiedSchemas) {
-        for (Map.Entry<String, Schema> componentSchema : schemas.entrySet()) {
-            Schema value = componentSchema.getValue();
-            updateSchemaWithReference(schemaName, modifiedName, value);
-            modifiedSchemas.put(modifiedName, value);
+        ParseOptions parseOptions = new ParseOptions();
+        SwaggerParseResult parseResult = new OpenAPIParser().readContents(openApiJson, null, parseOptions);
+        if (!parseResult.getMessages().isEmpty()) {
+            outStream.println("Sanitized OpenAPI contains errors:");
+            parseResult.getMessages().forEach(outStream::println);
```
TharmiganK, line 169: The logic is refactored to replace the string content of the OpenAPI model. For that reason, we have converted the OpenAPI model to string(in json format), do the replacing and convert it back to the OpenAPI model. This additionally ensures that the modified OpenAPI specification has no errors.

But here when we get any errors while parsing the modified OpenAPI json string to OpenAPI model, we print those errors and **anyway continue the code generation using the modified OpenAPI model**. Is that fine? Or should we fall-back to the original OpenAPI specification? If so we need to make a copy of the OpenAPI model before modifying. @daneshk @lnash94 FYI

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -116,137 +116,59 @@ public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi, Map<String, Strin
         return openapi;
     }
 
-    private static void modifyOASWithSchemaName(OpenAPI openapi, Map<String, String> nameMap) {
-        Components components = openapi.getComponents();
+    public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi) {
+        Optional<Map<String, String>> proposedNameMap = getProposedNameList(openapi);
```
ayeshLK, line 120: Any reason to return an `Map` wrapped with `Optional` here ? IMO if we have an empty collection, we should use either `Collections.emptyList` or `Collections.emptyMap` in those case (depending on the requirement).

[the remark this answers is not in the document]
TharmiganK: Yes, will change it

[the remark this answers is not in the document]
TharmiganK: Fixed with [11693ad](https://github.com/ballerina-platform/openapi-tools/pull/1766/commits/11693ad44baf485c02e35d48419e727fba710c02)

TharmiganK: Fixed with [11693ad](https://github.com/ballerina-platform/openapi-tools/pull/1766/commits/11693ad44baf485c02e35d48419e727fba710c02)

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -116,137 +116,59 @@ public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi, Map<String, Strin
...
+        String openApiJson = Json.pretty(openapi);
+        for (Map.Entry<String, String> entry : nameMap.entrySet()) {
+            String schemaName = entry.getKey();
+            String modifiedName = entry.getValue();
+            if (schemaName.equals(modifiedName)) {
+                continue;
             }
-        }
-    }
+            String patternString = ("(\\\"\\$ref\\\"\\s*:\\s*\\\"#/components/schemas/%s\\\")|" +
```
ayeshLK, line 156: Shall we extract this String template into a constant ? Um referring to the following part only;
```
"(\\\"\\$ref\\\"\\s*:\\s*\\\"#/components/schemas/%s\\\")|('\\$ref'\\s*:\\s*'#/components/schemas/%s')"
```

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -116,137 +116,59 @@ public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi, Map<String, Strin
...
-    private static void updateComponentRequestBodySection(Components components, String schemaName,
-                                                          String modifiedName,
-                                                          Map<String, RequestBody> modifiedRequestBodies) {
-        if (components.getRequestBodies() != null) {
-            for (Map.Entry<String, RequestBody> requestBody : components.getRequestBodies().entrySet()) {
-                RequestBody requestBodyValue = updateRequestBodySchemaType(schemaName, modifiedName,
-                        requestBody.getValue());
-                modifiedRequestBodies.put(requestBody.getKey(), requestBodyValue);
-            }
+            String replacement = "\"$ref\":\"#/components/schemas/%s\"".formatted(modifiedName).replace("$", "\\$");
```
ayeshLK, line 161: ditto

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -621,7 +408,7 @@ public static Map<String, String> collectParameterNames(List<Parameter> paramete
     }
 
     public static String replaceContentInBraces(String input, String expectedValue) {
-        String regex = "\\{([^}]*)\\}";
+        String regex = "\\{([^}]*)}";
```
ayeshLK, line 411: Lets mark this as a constant as well.

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
daneshk: What scenarios in which we end up with an error when converting back? Is it okay to continue code generation when there is an error in modified model?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
TharmiganK: But is it OK to provide the warning messages in a error stream?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
TharmiganK: Fixed via [21628fe](https://github.com/ballerina-platform/openapi-tools/pull/1766/commits/21628fe6c77a0f54d70984c8e48d6718612b1319)

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
TharmiganK: Fixed via [21628fe](https://github.com/ballerina-platform/openapi-tools/pull/1766/commits/21628fe6c77a0f54d70984c8e48d6718612b1319)

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
TharmiganK: Fixed via [21628fe](https://github.com/ballerina-platform/openapi-tools/pull/1766/commits/21628fe6c77a0f54d70984c8e48d6718612b1319)

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
daneshk: If the user can recover from the generation failure or compilation error by modifying the sanitized schema. I am +1 for continuing, as we give enough context to recover.

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
daneshk: As per [GNU C Library](https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html), the error stream is used for error messages and diagnostics issued by the program. So it is okay

TharmiganK: Ack, will change it to `System.err` 👍🏼

TharmiganK: Fixed with [21628fe](https://github.com/ballerina-platform/openapi-tools/pull/1766/commits/21628fe6c77a0f54d70984c8e48d6718612b1319)

openapi-cli/src/main/java/io/ballerina/openapi/cmd/BaseCmd.java
```diff
@@ -62,8 +62,7 @@ public class BaseCmd {
     @CommandLine.Option(names = {"--single-file"}, description = "Generate all contents in a single file")
     public boolean singleFile;
 
-    @CommandLine.Option(names = {"--use-sanitized-oas"}, hidden = true, description = "This is an experimental" +
-            " feature. This option enables code generation by modifying the given OAS to follow the" +
-            " Ballerina language best practices.")
+    @CommandLine.Option(names = {"--use-sanitized-oas"}, hidden = true, description = "This option enables code " +
```
lnash94, line 65: Are we going to add the subcommand `bal openapi sanitize` here?

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
lnash94: In the code generation process, we are going through the same parser issue check before reaching the modifier. Therefore, I prefer to continue code generation with the user-provided OAS when there is an issue after the modifier. Additionally, I'm in favor of adding a warning message to inform the user that we did not apply the Ballerina set of rules for the given OAS.

openapi-cli/src/main/java/io/ballerina/openapi/cmd/BaseCmd.java
TharmiganK: No, this PR is to make the existing implementation stable. Introducing `bal openapi sanitize` should be implemented separately on top of these changes

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -120,15 +122,16 @@ public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi, Map<String, Strin
         return openapi;
     }
 
-    public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi) {
+    public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi) throws BallerinaOpenApiException {
```
lnash94, line 125: Do we handle this exception where we are supposed to use this function?

TharmiganK: This is thrown from the generator functions: [`generateClientAndService`](https://github.com/ballerina-platform/openapi-tools/blob/3eda060f1933149b197714927fd5627c50a7399f/openapi-cli/src/main/java/io/ballerina/openapi/cmd/BallerinaCodeGenerator.java#L102)

In CLI, we capture this error, add a log and exit : https://github.com/ballerina-platform/openapi-tools/blob/3eda060f1933149b197714927fd5627c50a7399f/openapi-cli/src/main/java/io/ballerina/openapi/cmd/OpenApiCmd.java#L500

lnash94 (approved):

daneshk (approved):

PR #1771, opened 2024-09-12
Add OpenAPI mapping support for query annotation with name field

TharmiganK:
Fixes: https://github.com/ballerina-platform/ballerina-library/issues/6984

This PR maps the name from the `http:Query` annotation for the query parameter name in the generated OpenAPI specification. This new field is introduced via https://github.com/ballerina-platform/module-ballerina-http/pull/2143

### DIFF
--- a/docs/ballerina-to-oas/spec/spec.md
+++ b/docs/ballerina-to-oas/spec/spec.md
@@ -725,7 +725,7 @@ parameters:
       nullable: true
 ```
 
-The query parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using the `@http:Query` annotation.
+The query parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using the `name` attribute in the `@http:Query` annotation.
 
 ```ballerina
 service /api on new http:Listener(9090) {
@@ -856,7 +856,7 @@ parameters:
         nullable: true
 ```
 
-The header parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using the `@http:Header` annotation.
+The header parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using the `name` attribute in the `@http:Header` annotation.
 
 ```ballerina
 service /api on new http:Listener(9090) {
--- a/docs/ballerina-to-oas/spec/spec.md
+++ b/docs/ballerina-to-oas/spec/spec.md
@@ -725,6 +725,26 @@ parameters:
       nullable: true
 ```
 
+The query parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using the `@http:Query` annotation.
+
+```ballerina
+service /api on new http:Listener(9090) {
+
+  resource function get path(@http:Query{name: "userName"} string param) {
+      // ...
+  }
+}
+```
+
+```yml
+parameters:
+  - name: userName
+    in: query
+    required: true
+    schema:
+      type: string
+```
+
 If the query parameter type is a `map` or `record` with `anydata` fields, then the query parameter schema is wrapped with `content` and `application/json` to indicate that the query parameter should be a JSON object which should be encoded properly.
 
 ```ballerina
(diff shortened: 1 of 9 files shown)


docs/ballerina-to-oas/spec/spec.md
```diff
@@ -725,6 +725,26 @@ parameters:
...
+
+  resource function get path(@http:Query{name: "userName"} string param) {
+      // ...
+  }
+}
+```
+
+```yml
+parameters:
+  - name: userName
```
lnash94, line 741: It needs to be mentioned that `param` name details with `x-ballerina-name` in here, don't we?

TharmiganK: No, the extension is not related to ballerina to openapi specification.

docs/ballerina-to-oas/spec/spec.md
```diff
@@ -725,6 +725,26 @@ parameters:
       nullable: true
 ```
 
+The query parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using the `@http:Query` annotation.
```
NipunaRanasinghe, line 728:
```suggestion
The query parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using `name` attribute in the `@http:Query` annotation.
```

docs/ballerina-to-oas/spec/spec.md
```diff
@@ -836,6 +856,27 @@ parameters:
         nullable: true
 ```
 
+The header parameter name in the schema is defaulted to the Ballerina parameter name. But this can be overridden using the `@http:Header` annotation.
```
NipunaRanasinghe, line 859: ditto

docs/ballerina-to-oas/spec/spec.md
lnash94: Yeah, that makes sense, I see two options:

**Option 1**: We can include all modified and overridden names in the OAS. This way, when generating a service or client, we won’t need to worry about creating Ballerina-friendly names because we can simply pull them from the OAS.

**Option 2**: We keep the OAS clean by not adding any Ballerina-specific details, making it more universal and suitable for any programming language. Instead, we can append these extension details to our current implementation, which already handles mapping package specifics and includes Ballerina-specific details.

I prefer Option 2 to handle ballerina extension details with package detail implementations, WDYT?

docs/ballerina-to-oas/spec/spec.md
TharmiganK: Fixed via [9d88386](https://github.com/ballerina-platform/openapi-tools/pull/1771/commits/9d88386ba932a402494a8e0c04687a0a1f3aeb56)

docs/ballerina-to-oas/spec/spec.md
TharmiganK: Fixed via [9d88386](https://github.com/ballerina-platform/openapi-tools/pull/1771/commits/9d88386ba932a402494a8e0c04687a0a1f3aeb56)

docs/ballerina-to-oas/spec/spec.md
TharmiganK: Yes, I also prefer Option 2 to keep the OAS clean rather than adding Ballerina specific extensions. Users can always sanitize the OAS, if they want to generate Ballerina client or service.

lnash94: Shall we create a separate issue to track the extensions details adding? and let's keep our default behaviour as we fixed now.

openapi-cli/src/test/resources/ballerina-to-openapi/expected_gen/query/query_scenario10.yaml
```diff
@@ -0,0 +1,174 @@
...
+  /query:
+    get:
+      operationId: getQuery
+      parameters:
+        - name: query0
+          in: query
+          schema:
+            type: string
+            default: ""
+        - name: q1
```
lnash94, line 22: is this expected to have `q1` instead of `query1`?

TharmiganK: It should be `query1`, will fix it

lnash94 (approved):

NipunaRanasinghe (approved):

PR #1775, opened 2024-09-25
Decouple the flatten from the sanitize sub command

lnash94:
Introducing a separate flag for flattening will be addressed in future, and we will update the OpenAPI spec with the current default behaviour is flatten enabled. 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -114,9 +115,110 @@ public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi, Map<String, Strin
...
+    private void handleInlineObjectSchemaInComposedSchema(Schema<?> schema) {
+        if (schema instanceof ComposedSchema composedSchema) {
+            processSubSchemas(composedSchema.getAllOf());
+            processSubSchemas(composedSchema.getAnyOf());
+            processSubSchemas(composedSchema.getOneOf());
+        }
+    }
+
+    private void processSubSchemas(List<Schema> subSchemas) {
+        if (subSchemas != null) {
```
TharmiganK, line 178:
```suggestion
        if (Objects.nonNull(subSchemas)) {
```
Change in other places as well

openapi-cli/src/main/java/io/ballerina/openapi/cmd/Sanitize.java
```diff
@@ -54,7 +54,7 @@ public String getDefaultFileName() {
     @Override
     public Optional<OpenAPI> generate(String openAPIFileContent) {
         Optional<OpenAPI> filteredOpenAPI = getFilteredOpenAPI(openAPIFileContent);
-        return filteredOpenAPI.flatMap(this::getFlattenOpenAPI).flatMap(this::sanitizeOpenAPI);
+        return filteredOpenAPI.flatMap(this::sanitizeOpenAPI);
```
TharmiganK, line 57: Need to update the command description as well: https://github.com/ballerina-platform/openapi-tools/blob/3434ba64067c064f8263408b9b1c039dff90c6a0/openapi-cli/src/main/resources/cli-help/ballerina-openapi-sanitize.help#L15

openapi-cli/src/main/resources/cli-help/ballerina-openapi-sanitize.help
```diff
@@ -12,9 +12,7 @@ SYNOPSIS
 
 DESCRIPTION
        Sanitize the OpenAPI contract file according to the best naming
-       practices of Ballerina. The OpenAPI contract is flatten and the type
-       schema names are made Ballerina friendly. The Ballerina name
-       extensions are added to the schemas which can not be modified
+       practices of Ballerina. The Ballerina name extensions are added to the schemas which can not be modified
```
TharmiganK, line 15: There is a character limit when we print it in the console can we split it to the next line
```help
DESCRIPTION
       Sanitize the OpenAPI contract file according to the best naming
       practices of Ballerina. The Ballerina name extensions are added
       to the schemas which can not be modified directly.
```

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -114,9 +115,111 @@ public OpenAPI modifyWithBallerinaConventions(OpenAPI openapi, Map<String, Strin
         openapi = modifyOASWithSchemaName(openapi, nameMap);
         modifyOASWithParameterName(openapi);
         modifyOASWithObjectPropertyName(openapi);
+        modifyOASWithObjectPropertyNameInlineObjectSchema(openapi);
```
TharmiganK, line 118: Can't we combine this to the `modifyOASWithParameterName` function? That function already loop through the path parameters. So if we do this modification within the same visitor then it will efficient. Also I don't see a separate use case for this function, if we want to sanitize the parameters then we need to call both `modifyOASWithParameterName` and `modifyOASWithObjectPropertyNameInlineObjectSchema`

May be we can rename this function properly since the `modifyOASWithObjectPropertyNameInlineObjectSchema` sanitize the inline response schemas as well

lnash94: While implementing this had a same thought of combining handling inline object schema in parameter with this   `modifyOASWithParameterName`  function. avoiding combing this by assuming inline object schema is not in a common case, will check whether we can combing these two together.

lnash94: I checked the ability of combining, there are concerned with combing with the implementation `modifyOASWithParameterName`—this function only modifies parameter names.  `modifyOASWithObjectPropertyNameInlineObjectSchema` only modifies properties of object types.  If we incorporate these two, there can be additional unnecessary travers like, if there are no object types in the inline schema, the person would still need to go through the logic since it's tied to the parameter name modifier. 
Therefore , I thought to having both function separately and will modify the meaning full name, WDYT?

lnash94: Updated with https://github.com/ballerina-platform/openapi-tools/pull/1775/commits/b30ae39666d29f909a32bb60da3b06d4eb1ba1a3

TharmiganK (approved):

PR #1781, opened 2024-11-13
[2.1.x] Fix OpenAPI spec generation issues for U10 patch

lnash94:
1. Fixed included record field overriding issue https://github.com/wso2-enterprise/internal-support-ballerina/issues/831
2. Improvement: mapping constant values as default value in the generated OAS https://github.com/ballerina-platform/ballerina-library/issues/7353

## Automation tests
 - Unit tests - Added
   > Code coverage information
 - Integration tests

## Security checks


ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/parameter/AbstractParameterMapper.java
```diff
@@ -55,8 +59,15 @@ public void setParameter() throws ParameterMapperException {
         parameterList.forEach(operationInventory::setParameter);
     }
 
-    static Object getDefaultValue(DefaultableParameterNode parameterNode) {
-        Node defaultValueExpression = parameterNode.expression();
+    static Object getDefaultValue(DefaultableParameterNode parameterNode, SemanticModel semanticModel) {
+        ExpressionNode defaultValueExpression = (ExpressionNode) parameterNode.expression();
```
TharmiganK, line 63: Is it safe to cast here or do we have to check the type or kind?

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/parameter/AbstractParameterMapper.java
```diff
@@ -55,8 +59,15 @@ public void setParameter() throws ParameterMapperException {
...
+        Optional<Symbol> symbol = semanticModel.symbol(defaultValueExpression);
+        if (symbol.isPresent() && symbol.get() instanceof ConstantSymbol constantSymbol) {
+            Object constValue = constantSymbol.constValue();
+            if (constValue instanceof ConstantValue value) {
+                return value.value();
+            }
+        }
```
TharmiganK, line 70: Shall we separate this as a common util function to extract the constant value from node using the semantic model. It may be useful in the future

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/RecordTypeMapper.java
```diff
@@ -162,17 +172,19 @@ public static Map<String, Schema> mapRecordFields(Map<String, RecordFieldSymbol>
     }
 
     public static Optional<Object> getRecordFieldDefaultValue(String recordName, String fieldName,
-                                                    ModuleMemberVisitor moduleMemberVisitor) {
+                                                    ModuleMemberVisitor moduleMemberVisitor,
+                                                              SemanticModel semanticModel) {
```
TharmiganK, line 176: Seems like an alignment issue

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/RecordTypeMapper.java
```diff
@@ -183,6 +195,13 @@ private static Optional<Object> getRecordFieldDefaultValue(String fieldName,
             return Optional.empty();
         }
         ExpressionNode defaultValueExpression = defaultValueNode.expression();
+        Optional<Symbol> symbol = semanticModel.symbol(defaultValueExpression);
+        if (symbol.isPresent() && symbol.get() instanceof ConstantSymbol constantSymbol) {
+            Object constValue = constantSymbol.constValue();
+            if (constValue instanceof ConstantValue value) {
+                return Optional.of(value.value());
+            }
+        }
```
TharmiganK, line 204: Common util function will help here

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/parameter/AbstractParameterMapper.java
lnash94: Yeah, before adding this,  I checked with the lang team and, they informed that it is safe to cast because it returns expression nodes always.

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/RecordTypeMapper.java
```diff
@@ -120,7 +124,13 @@ static List<Schema> mapIncludedRecords(RecordTypeSymbol typeSymbol, Components c
                         .typeDescriptor();
                 Map<String, RecordFieldSymbol> includedRecordFieldMap = includedRecordTypeSymbol.fieldDescriptors();
                 for (Map.Entry<String, RecordFieldSymbol> includedRecordField : includedRecordFieldMap.entrySet()) {
-                    recordFieldMap.remove(includedRecordField.getKey());
+                    RecordFieldSymbol recordFieldSymbol = recordFieldMap.get(includedRecordField.getKey());
+                    RecordFieldSymbol includedRecordFieldValue = includedRecordField.getValue();
+                    boolean isRemovableField = recordFieldSymbol != null && includedRecordFieldValue.typeDescriptor()
+                            .equals(recordFieldSymbol.typeDescriptor()) && !recordFieldSymbol.hasDefaultValue();
```
TharmiganK, line 130: This logic should incorporate the optionality as well. For example:
```bal
type RecA record{|
     string a?;
|};

// The schema should mark a as a required field
type RecB record{|
     *RecA;
     string a;
     int b;
|};
```

I think we need to check whether both has the same result when we check for optionality

lnash94: Yeah you are right, This logic doesn’t handle the optional behavior. I left it unhandled because if we can map the scenarios mentioned in the comment[1], we could adjust the required section in the object.

However, in cases where `a` field is required in `RecA `but not in `RecB`[2], this can’t be represented accurately with `allOf`. Here, we have to consolidate all fields into a single record `RecB` (RecB object Schema), like this[3]:

[1] where `a` field is opetional in `RecA `but required  in `RecB`
```ballerina
type RecA record{|
     string a?;
|};

type RecB record{|
     *RecA;
     string a;
     int b;
|};
```
[2] where `a` field is required in `RecA `but not in `RecB`
```ballerina
type RecA record{|
     string a;
|};

type RecB record{|
     *RecA;
     string a?;
     int b;
|};
```
[3] oas representation
```yml
RecB:
   type: object
   required:
    - b
   properties:
       a:
        type: string
       b:
        type: integer
```
In addition, to support scenario [1], we need to add support OAS-to-Ballerina type generation also for correct end-to-end flow. In the second scenario[2], when regenerating the type with a specific mapping, we couldn’t capture `RecA` details within `RecB`. 

These are the reasons I’ve omitted the optional check in this phase of the fix, as we can still achieve a reasonably accurate representation for scenario 1, with only the required details missing.

I plan to address this optional behavior in a separate issue, as we need to handle both directions: OAS to Ballerina and Ballerina to OAS. WDYT?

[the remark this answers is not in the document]
lnash94: Updated the logic to check the default values

TharmiganK: I think we already support scenario [1] in OAS to Ballerina generation:
```yaml
RecA:
  type: object
  properties:
    a:
      type: string
RecB:
  type: object
  allOf:
  - $ref: "#/components/schemas/RecA"
  - type: object
    required:
    - a
    - b
    properties:
      a:
        type: string
      b:
        type: integer
        format: int64
```
This should generate the following types:
```bal
type RecA record{|
     string a?;
|};

type RecB record{|
     *RecA;
     string a;
     int b;
|};
```

Regarding scenario [2], I also agree that it is not a valid case to use `allOf`. So there should be validation happen before using `allOf` schema here(narrowing should be supported but expanding should not be allowed). But I think this not a 80% case in OpenAPI spec or in Ballerina code. Even for this scenario, when we generate the following incorrect schema(the meaning is different when you look at the schema without the context of Ballerina type inclusion) it should work in the code generation.
```yaml
RecA:
  type: object
  required:
  - a
  properties:
    a:
      type: string
RecB:
  type: object
  allOf:
  - $ref: "#/components/schemas/RecA"
  - type: object
    required:
    - b
    properties:
      a:
        type: string
      b:
        type: integer
        format: int64
```

My concern is that the scenario one is a mostly used one. So we should at least support that.

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/RecordTypeMapper.java
```diff
@@ -126,17 +126,81 @@ static List<Schema> mapIncludedRecords(RecordTypeSymbol typeSymbol, Components c
                 for (Map.Entry<String, RecordFieldSymbol> includedRecordField : includedRecordFieldMap.entrySet()) {
                     RecordFieldSymbol recordFieldSymbol = recordFieldMap.get(includedRecordField.getKey());
                     RecordFieldSymbol includedRecordFieldValue = includedRecordField.getValue();
-                    boolean isRemovableField = recordFieldSymbol != null && includedRecordFieldValue.typeDescriptor()
-                            .equals(recordFieldSymbol.typeDescriptor()) && !recordFieldSymbol.hasDefaultValue();
-                    if (isRemovableField) {
-                        recordFieldMap.remove(includedRecordField.getKey());
+
+                    if (recordFieldSymbol == null
```
TharmiganK, line 130: Can't we use `hasKey` first rather than calling `get` directly

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/RecordTypeMapper.java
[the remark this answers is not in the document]
TharmiganK: The logic will only work for the types in the same package and will not work for the types imported from a different package. In that case, we can ignore this logic and add the default value field again rather than checking the default values. I have suggested this to check whether there are any APIs from semantic model to get the fields defined in a record rather than getting all the effective fields (fields defined in record and fields coming from included record)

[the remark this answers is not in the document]
MaryamZi: I don't think there is. cc @dulajdilshan 

Also I'm not entirely sure about this check. Wouldn't this handle the following incorrectly?

```ballerina
type RecA record {|
      string a = "default a";
      string b;
|};

type RecB record {|
      *RecA;
      string a; // makes a required with no default
      int d;
|};
```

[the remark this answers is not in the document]
lnash94: @MaryamZi , This is handled correctly with this logic update: https://github.com/ballerina-platform/openapi-tools/pull/1781#discussion_r1841059023

Generated OAS for example,

```
RecA:
      required:
      - b
      type: object
      properties:
        a:
          type: string
          default: default a
        b:
          type: string
      additionalProperties: false
    RecB:
      type: object
      allOf:
      - $ref: '#/components/schemas/RecA'
      - required:
        - a
        - d
        type: object
        properties:
          a:
            type: string
          d:
            type: integer
            format: int64
        additionalProperties: false
```
I will update the test cases

@TharmiganK, AFAIK in the semantic model, we receive a typeSymbol, and that typeSymbol only contains the API for checking whether it has a default value. However, we cannot obtain the default value information from it. This was discussed during the meeting some time ago. @dulajdilshan will be able to verify if they know have an API we can use for getting value?

lnash94: There are two ways we can map scenario one in OAS

ballerina scenario: 
```ballerina
type RecA record{|
     string a?;
|};

type RecB record{|
     *RecA;
     string a;
     int b;
|};
```
- [1] way of mapping oas
```openapi
RecA:
  type: object
  properties:
    a:
      type: string
RecB:
  type: object
  allOf:
  - $ref: "#/components/schemas/RecA"
  - type: object
    required:
    - a
    - b
    properties:
      a:
        type: string
      b:
        type: integer
        format: int64
```

- [2] way of mapping to oas
```openapi
RecA:
  type: object
  properties:
    a:
      type: string
RecB:
  type: object
  allOf:
  - $ref: "#/components/schemas/RecA"
  - type: object
    required:
    - a
    - b
    properties:
      b:
        type: integer
        format: int64
```
when comes to the comparison in the usage of OAS spec, The [2] definition appears to be more accurate, as it uses [`allOf` effectively](https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/#:~:text=%3A%203%20%7D-,allOf,-OpenAPI%20lets%20you) to extend `RecA` without redefining the `a` property in `RecB`. This approach enables `RecB` to inherit the `a` property from `RecA`, which helps avoid redundancy and potential conflicts.

[My first comment](https://github.com/ballerina-platform/openapi-tools/pull/1781#discussion_r1840694203) I was discussing the scenario of mapping into the [2] OAS definition. The issue here is that our OAS-to-Ballerina code generation doesn’t yet support this specific definition, which is why we need support in both directions. On the other hand, yeah we already have support for the [1] definition in the tool.

I think it would be great if we could enable code generation for Ballerina to OAS using the [2] definition since our tool is designed to align with best practices while mapping definitions.

That said, I’ll add optional support for OAS to Ballerina code generation with the [2] definition included.

As we discussed earlier, the scenario **where `a` field is required in RecA but not in RecB** won’t be addressed here, as it falls outside the main 80% use cases we’re targeting.

 
**Summary: What we support from this fix in optional behaviour ;**

1. Support where `a`  field is optional in `RecA` but required in `RecB`

```ballerina
type RecA record{|
     string a?;
|};

type RecB record{|
     *RecA;
     string a;
     int b;
|};
```
Generated OAS:
```openapi
RecA:
  type: object
  properties:
    a:
      type: string
RecB:
  type: object
  allOf:
  - $ref: "#/components/schemas/RecA"
  - type: object
    required:
    - a
    - b
    properties:
      b:
        type: integer
        format: int64
```
2.  OAS to Ballerina Code generation both [1] and [2] will map to Ballerina 

```ballerina
type RecA record{|
     string a?;
|};

type RecB record{|
     *RecA;
     string a;
     int b;
|};
```
2. Not support: where a field is required in RecA but not in RecB
```ballerina
type RecA record{|
     string a;
|};

type RecB record{|
     *RecA;
     string a?;
     int b;
|};
```

lnash94: Fixed with these commit
- [Add optional field mapper](https://github.com/ballerina-platform/openapi-tools/pull/1781/commits/764d7bf4807712da7583b8ec72f70d3a911b0e7e)
- [Add optional field mapper for OAS to Ballerina code generation](https://github.com/ballerina-platform/openapi-tools/pull/1781/commits/810d9a49176eda1d23d7a028578b11793af2fe2a)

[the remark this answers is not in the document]
dulajdilshan: > However, we cannot obtain the default value information from it.

Semantic API doesn't support getting default value information regarding fields in the records.

lnash94: with this fix defualt value in separate package record fields are not captured.
If we have a record defined in a separate package, for example, `packageA`:

```ballerina
type RecA record {|
    int a = 10;
    string c = "c";
|};

```
and it is included in `packageB`:

```ballerina

type RecB record {|
    *packageA:RecA;
    string b = "c";
|};
```
there is a limitation in the OpenAPI Specification (OAS) mapping for RecB. This limitation arises because the default value access API cannot retrieve default values from another package. As a result, we generate all the fields explicitly for RecB to ensure accessibility.

Generated yaml
```yml
RecA:
      type: object
      properties:
        a:
          type: integer
          default: 10
        c:
          type: string
          default: c
      additionalProperties: false
    RecB:
      type: object
      allOf:
      - $ref: '#/components/schemas/RecA'
      - required:
        - b
        type: object
        properties:
          b:
            type: integer
            format: int64
          a:
            type: integer
          c:
            type: string
          statusCodes:
            type: array
            items:
              type: integer
              format: int64
        additionalProperties: false
    ```

TharmiganK (approved):
LGTM

HindujaB:
This fix will be released with 2201.10.3 patch
https://github.com/ballerina-platform/ballerina-distribution/issues/5823

PR #1821, opened 2025-02-19
[Update 12] Introduce a new API to convert a Ballerina Type Symbol to a Json Schema(String Format)

TharmiganK:
## Purpose

> $Subject

> **Note:** This is only supported for the `anydata` types, for other types an unsupported operation exception will be thrown.

Additionally, we have supported enable expansion option which will replace the definition references by inline schemas which will make the definition obsolete. Please note that this will only work for types which does not have cyclic references.

### DIFF
--- a/ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapper.java
+++ b/ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapper.java
@@ -56,4 +56,6 @@ public interface TypeMapper {
         }
         schema.setDefault(defaultValue);
     }
+
+    String getJsonSchemaString(TypeSymbol typeSymbol);
 }
--- a/ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapperImpl.java
+++ b/ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapperImpl.java
@@ -17,6 +17,7 @@
  */
 package io.ballerina.openapi.service.mapper.type;
 
+import io.ballerina.compiler.api.SemanticModel;
 import io.ballerina.compiler.api.symbols.ArrayTypeSymbol;
 import io.ballerina.compiler.api.symbols.ErrorTypeSymbol;
 import io.ballerina.compiler.api.symbols.IntersectionTypeSymbol;
@@ -29,13 +30,22 @@ import io.ballerina.compiler.api.symbols.TypeReferenceTypeSymbol;
 import io.ballerina.compiler.api.symbols.TypeSymbol;
 import io.ballerina.compiler.api.symbols.UnionTypeSymbol;
 import io.ballerina.openapi.service.mapper.model.AdditionalData;
+import io.ballerina.openapi.service.mapper.model.ModuleMemberVisitor;
+import io.ballerina.openapi.service.mapper.type.extension.BallerinaTypeExtensioner;
+import io.ballerina.projects.Project;
+import io.ballerina.projects.plugins.SyntaxNodeAnalysisContext;
+import io.swagger.v3.core.util.Json;
+import io.swagger.v3.core.util.OpenAPISchema2JsonSchema;
 import io.swagger.v3.oas.models.Components;
 import io.swagger.v3.oas.models.media.Schema;
 
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
+import static io.ballerina.openapi.service.mapper.ServiceToOpenAPIMapper.extractNodesFromProject;
 import static io.ballerina.openapi.service.mapper.utils.MapperCommonUtils.getTypeName;
 
 /**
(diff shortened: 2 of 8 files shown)


ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/SchemaResult.java
```diff
@@ -0,0 +1,31 @@
+/*
+ *  Copyright (c) 2024, WSO2 LLC. (http://www.wso2.com).
```
ayeshLK, line 2:
```suggestion
 *  Copyright (c) 2025, WSO2 LLC. (http://www.wso2.com).
```

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapperImpl.java
```diff
@@ -46,14 +61,97 @@
...
+        Project project = context.currentPackage().project();
+        ModuleMemberVisitor moduleMemberVisitor = extractNodesFromProject(project, semanticModel);
+        this.componentMapperData = new AdditionalData(semanticModel, moduleMemberVisitor);
+    }
+
+    public Schema getSchema(TypeSymbol typeSymbol) throws UnsupportedOperationException {
+        return getSchema(typeSymbol, true).schema();
+    }
+
+    public SchemaResult getSchema(TypeSymbol typeSymbol, boolean enableExpansion)
```
ayeshLK, line 90: It seems like this API is used internally. Do we need to make it public in that case ?

TharmiganK: If we want to get the `SchemaResult`(type schema and the relevant components) by disabling the `enableExpansion`, we need this API

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapperImpl.java
```diff
@@ -46,14 +61,97 @@
  */
 public class TypeMapperImpl implements TypeMapper {
 
+    public static final String COMPONENTS_SCHEMAS = "#/components/schemas/";
+    public static final String DEFINITIONS = "#/definitions/";
+    public static final String DEFINITIONS_FIELD = "definitions";
+    public static final String JSON_SCHEMA_FIELD = "$schema";
+    public static final String JSON_SCHEMA_VERSION = "http://json-schema.org/draft-04/schema#";
```
lnash94, line 68: Is there compatibility for OAS v3.1 with the JSON Schema draft version? The previous mention was for OAS 3.0 with JSON Draft-05 support and the Draft 4 meta-schemas.

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapperImpl.java
```diff
@@ -46,14 +61,97 @@
  */
 public class TypeMapperImpl implements TypeMapper {
 
+    public static final String COMPONENTS_SCHEMAS = "#/components/schemas/";
+    public static final String DEFINITIONS = "#/definitions/";
```
lnash94, line 65: Are we going to generate an OAS with the swagger (OpenAPI 2) version as well?

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapper.java
```diff
@@ -56,4 +56,12 @@ static void setDefaultValue(Schema schema, Object defaultValue) {
         }
         schema.setDefault(defaultValue);
     }
+
+    SchemaResult getSchema(TypeSymbol typeSymbol, boolean enableExpansion) throws UnsupportedOperationException;
+
+    Schema getSchema(TypeSymbol typeSymbol) throws UnsupportedOperationException;
+
+    String getJsonSchemaString(TypeSymbol typeSymbol) throws UnsupportedOperationException;
```
lnash94, line 64: Shall we add tests for this new API?

TharmiganK: At the moment we don't have unit tests for this package. Thinking of addressing this in a separate issue.

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapperImpl.java
TharmiganK: I could not found an information regarding this. This is added by considering the minimal support. Also we only adding specific fields from our side, so it should be compatible with the latest version as well. (Btw the actual schema is not an optimised Json Schema, as mentioned in the original issue, the JSON specific schema generation should be supported from jsondata package. And here we only provide an API as a workaround)

lnash94: Yeah, This is fine because we're supporting the initial stage of this effort as a workaround, and +1  to the capability come from the `jsondata` package.

ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/TypeMapper.java
TharmiganK: Writing a compiler plugin just to test this API seems to be not efficient approach. Since this is for an internal requirement and the package which uses this API has tests, we can ignore the testing part for now.

@MohamedSabthar we have tests from the agent library right?

MohamedSabthar: Yes, we are currently adding few test cases.

lnash94 (approved):

PR #1836, opened 2025-05-16
Fix common path extraction logic

TharmiganK:

### DIFF
--- a/openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
+++ b/openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
@@ -973,7 +973,7 @@ public class OASModifier {
                 break;
             }
         }
-        return (commonPath.isEmpty() || commonPath.trim().equals(SLASH)) ? Optional.empty() : Optional.of(commonPath);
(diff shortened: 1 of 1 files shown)


daneshk (approved):

openapi-core/src/main/java/io/ballerina/openapi/core/generators/common/OASModifier.java
```diff
@@ -973,7 +973,7 @@ private static Optional<String> getCommonPath(List<String> pathValues) {
                 break;
             }
         }
-        return commonPath.isEmpty() ? Optional.empty() : Optional.of(commonPath);
+        return (commonPath.isEmpty() || commonPath.trim().equals(SLASH)) ? Optional.empty() : Optional.of(commonPath);
```
niveathika, line 976: Can commonPath have just spaces? In this case this will not return empty

TharmiganK: Generally the path segments will not have a space. But still it is a valid path. In that case the common path should be with those spaces. So we just need to check whether commonPath is empty or just a slash:
```suggestion
        return (commonPath.isEmpty() || commonPath.equals(SLASH)) ? Optional.empty() : Optional.of(commonPath);
```

daneshk (approved):

PR #1837, opened 2025-05-16
Fix issues with flatten command

TharmiganK:
## Purpose

Fixes the following issue:
- Generated reference name with dot is making the OpenAPI spec invalid
- Flattening properties, request body and response should keep the description. Other wise there is no documentations for those fields. => This is fixed in the generation where we get description by resolving the references
- Remove default servers generated for the specifications which does not have a servers section


SachinAkash01 (approved):
LGTM

daneshk (approved):

PR #1846, opened 2025-09-11
[2201.12.x] Improve server port extraction to support port values provided from module level variables

TharmiganK:
## Purpose

> Fixes: https://github.com/ballerina-platform/ballerina-library/issues/8219

Other bug fixes:
- https://github.com/ballerina-platform/ballerina-library/issues/8240
- https://github.com/ballerina-platform/ballerina-library/issues/7687
- https://github.com/ballerina-platform/ballerina-library/issues/7679
- https://github.com/ballerina-platform/ballerina-library/issues/8033

**Before this change:**
- When we cannot extract the port value, the tool always stick to a default value (The tool spec says it should use 80 for http and 443 for https. But with the default listener change, we change it to 9090)

**After this change:** [Edited according to this [comment](https://github.com/ballerina-platform/openapi-tools/pull/1846#issuecomment-3279799753)]
- If the listener is the http default listener then we will default it to 9090. This a reasonable assumption we make, but still the port can be changed using a configurable file which is not recommended
- If the listener port is a literal value or a module level variable which ultimately has a literal value, then the servers information will have the extracted port value. Since the variables are in module level and we are using service declarations to define the services, this will give the correct port at runtime also.
- If the listener port is a configurable variable, then we use the default value of the variable as a reasonable value, but still the port can be changed using a configurable file. If the configurable variable does not have a default value i.e. if it is a required configurable variable then we provide a warning and the `servers` information will not be in the generated specification
- For any other expressions for the port will be considered as unsupported(will have proper warning messages) and an **error will be thrown**

### DIFF
--- a/ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/SimpleTypeMapper.java
+++ b/ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/SimpleTypeMapper.java
@@ -107,10 +107,9 @@ public class SimpleTypeMapper extends AbstractTypeMapper {
             case ANYDATA:
                 return new Schema();
             case NEVER:
-                // TODO: Check whether this can be represented with `not` in object schema
-                // Current behaviour will be skipping the type mapping
-                return null;
             default:
+                // TODO: For `NEVER`, check whether this can be represented with `not` in object schema
+                // Current behaviour will be skipping the type mapping
                 return null;
         }
     }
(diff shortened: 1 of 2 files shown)


ballerina-to-openapi/src/main/java/io/ballerina/openapi/service/mapper/type/SimpleTypeMapper.java
```diff
@@ -105,6 +106,10 @@ public static Schema getSchema(TypeDescKind typeKind) {
                 return schema;
             case ANYDATA:
                 return new Schema();
+            case NEVER:
+                // TODO: Check whether this can be represented with `not` in object schema
+                // Current behaviour will be skipping the type mapping
+                return null;
             default:
                 return null;
```
lnash94, line 114:
```suggestion
            case NEVER:
            default:
                // TODO: For `NEVER`, check whether this can be represented with `not` in object schema
                // Current behaviour will be skipping the type mapping
                return null;
```

